PR #8, opened 2017-11-24
Add assets support to django container

bartoszbetka:
Add assets support to django container. Waiting for review.

### DIFF
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -18,7 +18,7 @@ $(info CONCENT_DIR=$(CONCENT_DIR))
 
 all: build/ nginx-proxy concent-api
 
-nginx-proxy: $(NGINX_PROXY_SOURCE)
+nginx-proxy: $(NGINX_PROXY_SOURCE) build/nginx-proxy/concent-api-assets.tar
 	docker build -t nginx-proxy:$(VERSION) build/nginx-proxy/
 	docker tag nginx-proxy:$(VERSION) nginx-proxy:latest
 
@@ -39,6 +39,13 @@ build/concent-api/concent-api.tar: $(CONCENT_DIR)
 	mkdir --parents $(dir $@)
 	package-builder/build-concent-api.sh "$(CONCENT_DIR)" "build/concent-api"
 
+build/nginx-proxy/concent-api-assets.tar: $(CONCENT_DIR)
+	mkdir --parents "$(dir $@)"
+	package-builder/build-concent-api-assets.sh \
+		"$(CONCENT_DIR)"                        \
+		"$(abspath package-builder)"            \
+		"$(abspath build)"                      \
+		"$(dir $@)"
 
 build/:
 	mkdir --parents "$@"
--- a/containers/package-builder/build-concent-api-assets.sh
+++ b/containers/package-builder/build-concent-api-assets.sh
@@ -1,0 +1,25 @@
+#!/bin/bash -e
+
+concent_dir=$(realpath "$1")
+package_builder_dir=$(realpath "$2")
+build_dir=$(realpath "$3")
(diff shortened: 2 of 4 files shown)


containers/package-builder/build-concent-api-assets.sh
```diff
@@ -0,0 +1,23 @@
...
+output_dir=$2
+
+mkdir --parents build/concent_api_assets/
+cp --recursive $concent_dir/concent_api build/concent_api_assets
+
+cd build/concent_api_assets/concent_api/
+
+virtualenv --python python3 concent-api-virtualenv/
+source concent-api-virtualenv/bin/activate
+pip install -r "requirements.txt"
```
cameel: This fails on my machine:

```
package-builder/build-concent-api-assets.sh: /home/cameel/data/working-set/projects/golem/concent-deployment/containers/build/concent_api_assets/concent_api/concent-api-virtualenv/bin/pip: /home/cameel/data/working-set/projects/golem/concent-deployment/containers/build/concent_api_assets/concent_api/concent-api-v: bad interpreter: No such file or directory
make: *** [Makefile:44: build/nginx-proxy/concent-api-assets.tar] Error 126
```

Turns out that the length of the shebang line is limited to 127 characters in Linux (see [The #! magic, details about the shebang/hash-bang mechanism on various Unix flavours > Further history and the maximum length of a #! line](https://www.in-ulm.de/~mascheck/various/shebang/#length)). `virtualenv` embeds the full path to the python executable in all scripts in the `/bin/` directory, including `pip`. If you open the `pip` script, the first line looks like this:

``` python
#!/home/cameel/data/working-set/projects/golem/concent-deployment/containers/build/concent_api_assets/concent_api/concent-api-virtualenv/bin/python3
```

This is too long and gets cut off.

There are various workarounds but in our case the simplest one is to simply put the virtualenv in the `build/` directory rather than deeper, inside `concent-api-virtualenv/`.

containers/Makefile
```diff
@@ -39,6 +39,9 @@ build/concent-api/concent-api.tar: $(CONCENT_DIR)
 	mkdir --parents $(dir $@)
 	package-builder/build-concent-api.sh "$(CONCENT_DIR)" "build/concent-api"
 
+build/nginx-proxy/concent-api-assets.tar: $(CONCENT_DIR)
+	mkdir --parents $(dir $@)
+	package-builder/build-concent-api-assets.sh "$(CONCENT_DIR)" "../../nginx-proxy"
```
cameel: `../../nginx-proxy` here is relative to the `build/concent-api-assets/concent_api` path we `cd` into inside the script, which, in turn is relative to the directory of the `Makefile`. This is different from other paths in the `Makefile` and very confusing for someone reading the script and will surely be a source of future bugs.

You should always assume that the paths are relative to script's working directory. If you need something relative to another directory, just use the absolute path.

In bash, use `$(realpath "dir")` to get the absolute path. In make use `$(abspath dir)`

containers/Makefile
```diff
@@ -16,9 +16,9 @@ ifndef CONCENT_DIR
 endif
 $(info CONCENT_DIR=$(CONCENT_DIR))
 
-all: build/ nginx-proxy concent-api
+all: build/ concent-api nginx-proxy
 
-nginx-proxy: $(NGINX_PROXY_SOURCE)
+nginx-proxy: $(NGINX_PROXY_SOURCE) concent-api build/nginx-proxy/concent-api-assets.tar
```
cameel: There's no reason for `nginx-proxy` docker image to depend on the `concent-api` image. One does not use the other.

cameel:
I have fixed all the issues and merged the branch into master.

Issue #15, opened 2017-11-30
nginx-storage pod with persistent disk


cameel:
- Create a Kubernetes deployment and service for `nginx-storage` - an instance of nginx that we'll use for file uploads and downloads.
    - The service should **not** have an external IP (for now).
- Provision 30 GB of storage and attach it as a volume to the container.
    - This size is for development only. In production we'll want to have more storage.
    - Use standard persistent storage, not SSD.
    - Just like cluster creation, buying storage should **not** be done by our deployment playbooks. Do it manually and describe the `gcloud` commands to do it in the README. In deployment playbooks just assume that the storage already exists under a predefined name (that should be defined in [concent-deployment-values](https://github.com/golemfactory/concent-deployment-values)).
- Configure `nginx-proxy` to pass all HTTP requests for `/download/` and `/upload/` to `nginx-storage`.

cameel:
Done in #20 and merged into master.

PR #20, opened 2017-12-05
 nginx-storage pod with persistent disk

bartoszbetka:
#15

### DIFF
--- a/kubernetes/Makefile
+++ b/kubernetes/Makefile
@@ -1,6 +1,7 @@
 CLUSTER_SCRIPTS :=                                     \
 	build/services/nginx-proxy.yml                     \
 	build/services/concent-api.yml                     \
+	build/services/nginx-storage.yml                   \
 	build/create-job-dependencies.sh                   \
 	build/delete-job-dependencies.sh                   \
 	build/clean-up-jobs.sh                             \
@@ -11,6 +12,7 @@ CLUSTER_SCRIPTS :=                                     \
 	build/push-docker-images.sh                        \
 	build/config-maps/nginx-proxy/default.conf         \
 	build/config-maps/concent-api/local_settings.py    \
+	build/config-maps/nginx-storage/default.conf       \
 	build/concent-secrets/secrets.py                   \
 	build/secrets/db-secrets.yml                       \
 	build/secrets/django-admin-fixture.yaml            \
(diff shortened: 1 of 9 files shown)


kubernetes/Makefile
```diff
@@ -16,7 +16,9 @@ CLUSTER_SCRIPTS :=                                     \
 	build/secrets/django-admin-fixture.yaml            \
 	build/jobs/create-database.yml                     \
 	build/jobs/migrate-database.yml                    \
-	build/jobs/delete-database.yml
+	build/jobs/delete-database.yml                     \
+	build/config-maps/nginx-storage/default.conf       \
+	build/services/nginx-storage.yml
```
cameel: Please keep similar files together. This should be next to the other files in `build/services/`.

kubernetes/services/nginx-storage.yml.j2
```diff
@@ -0,0 +1,53 @@
...
+                command: ["/usr/sbin/nginx", "-s", "quit"]
+          volumeMounts:
+            - mountPath:  /srv/storage/
+              name:       concent-storage
+            - mountPath:  /etc/nginx/conf.d/
+              name:       nginx-storage-config
+      volumes:
+        - name: concent-storage
+          gcePersistentDisk:
+            pdName: {{ persistence_disk_name }}
```
cameel: `persistence` > `persistent`.

Also, "persistent disk" is a very generic name and does not explain well how this disk is different from others persistent disks we use. For example every cluster node has a persistent disk for its files attached. Call it `concent_storage_disk_name`.

cameel:
@bartoszbetka Please fix the issues listed above so that we can merge the branch.

bartoszbetka:
Issues fixed.

cameel:
Code merged. I made a few small tweaks myself:
- Rewritten the snippet in the README
- Your path in `nginx-proxy` was missing the final slash.
- Changed the order in which pods are created and deleted. `nginx-proxy` should always be created last and deleted first because it's the entry point to the cluster. Similar with `nginx-storage` (but it's not in a cluster of its own yet so it's after `nginx-proxy`). Then everything else.
- I saw that `delete-services.sh` still does not ignore errors from `delete-config-maps.sh` which it calls. It was supposed to be fixed by now.

PR #24, opened 2017-12-15
Nginx storage upload

### LINKED ISSUE #17: Uploading files to nginx-storage without authentication
Configure `nginx-storage` (#15) to allow uploading arbitrary files to its persistent disk. At this point accept anything, without worrying about permissions - this will be the subject of another task.

A POST request to `/upload/` with `multipart/form-data` content type and `Concent-Upload-Path` HTTP header should result in the included file being placed under the name specified in that header on the persistent disk. The name may be a relative path. The name specified in `Content-Disposition` of the multipart request should be ignored.

Allow uploading only one file at a time. A request with zero or more than one file should result in HTTP 400 error.

If possible, try to prevent overwriting existing files. If a file already exists, return a 4xx error. But if not create a separate task for this. It's something we'll have to deal with in a different way.

Ensure that if a download is interrupted, the partial file is not left under the target name (but it's OK to keep such files in temporary location and allow resuming uploads).

Enforce limits both on the size of individual files and on the total size of uploaded files. Limits should be configurable in [concent-deployment-values](https://github.com/golemfactory/concent-deployment-values).

### Example request
``` bash
curl -X POST http://storage.concent.golem.network/upload/ \
    --header "Concent-Upload-Path: blender/benchmark/test_task/scene-Helicopter-27-cycles.blend" \
    --form file=@scene-Helicopter-27-cycles.blend
```

bartoszbetka:
Resolves #17


kubernetes/services/nginx-storage.yml.j2
```diff
@@ -11,7 +11,7 @@ spec:
         run: nginx-storage
     spec:
       containers:
-        - image:           nginx:1.13.6-alpine
+        - image:           {{ docker_registry }}/nginx-storage:latest
```
cameel: I have merged @Gr33ny's branch that adds `image_prefix` and `image_version` today. Please add them here and rebase your branch on master.

kubernetes/services/nginx-storage.yml.j2
```diff
@@ -38,6 +38,8 @@ spec:
         - name: nginx-storage-config
           configMap:
             name: nginx-storage-config
+      securityContext:
+        fsGroup: 65534
```
cameel: This needs some explanation (in comments). Why are you using `fsGroup` and why this particular value? Does this require the container to be built or configured in a specific way?

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,96 @@
...
+
+    location = /upload/ {
+
+        access_by_lua_block {
+            -- Module nginx big upload support only content-type: "application/x-www-form-urlencoded"
+            check_content_type_header=ngx.req.get_headers()["Content-Type"]
+            if check_content_type_header ~= "application/x-www-form-urlencoded" then
+                ngx.status=400
+                ngx.say('Only Content-Type: "application/x-www-form-urlencoded" is allowed!')
+                ngx.exit(400)
```
cameel: Why both `ngx.status=400` and `ngx.exit(400)`? Which one is the response status and what does the other do?

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,96 @@
...
+
+    location /download/ {
+        alias /srv/storage/;
+    }
+
+    location = /upload/ {
+
+        access_by_lua_block {
+            -- Module nginx big upload support only content-type: "application/x-www-form-urlencoded"
+            check_content_type_header=ngx.req.get_headers()["Content-Type"]
```
cameel: Please don't call variables like this. If you start the name with a verb, it sounds like a function. Should be `content_type_header`.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,96 @@
...
+
+
+    location /download/ {
+        alias /srv/storage/;
+    }
+
+    location = /upload/ {
+
+        access_by_lua_block {
+            -- Module nginx big upload support only content-type: "application/x-www-form-urlencoded"
```
cameel: Does this module really suppport `application/x-www-form-urlencoded`? Or does it simply ignore the content type?

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,96 @@
...
+            end
+        }
+
+        proxy_pass http://nginx-storage/upload_internal/;
+        # Make sure that user could not submitted header "X-Session-Id" with content
+        # This header is responsible for temporary filename
+        proxy_set_header X-Session-Id "";
+    }
+
+    location /upload_internal/ {
```
cameel: Our convention is to use dashes rather than underscores in URLs.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: The right response in this case is `HTTP 415 UNSUPPORTED MEDIA TYPE`.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: Why is there no `internal` directive inside  `/upload-internal/`? It should not be accessible from the outside.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,96 @@
...
+        lua_code_cache on;
+        set $storage backend_file;
+        set $bu_checksum on;
```
cameel: What does `bu_checksum` do? Why do we need it?

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,96 @@
...
+    gzip                     on;
+    gzip_min_length          1000;
+    gzip_proxied             expired no-cache no-store private auth;
+    gzip_types               text/plain text/css text/javascript application/xml;
+    client_max_body_size     {{ max_upload_size }};
+
+
+    location /download/ {
+        alias /srv/storage/;
+    }
```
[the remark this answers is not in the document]
bartoszbetka: This code was deleted from nginx-storage-upload branch.

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: Not working with `internal` directive. In nginx documentation, you can see possible internal request. http://nginx.org/en/docs/http/ngx_http_core_module.html#internal.

containers/nginx-storage/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openresty/openresty:alpine-fat
+
+COPY install-dependencies.sh /tmp/
+RUN /tmp/install-dependencies.sh
+
+COPY table_length.lua        /usr/local/openresty/lualib/
+CMD ["/usr/local/openresty/bin/openresty", "-g", "daemon off;"]
```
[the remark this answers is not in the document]
bartoszbetka: I don't find that option `-g ` have long name. Include `-g daemon off;` in dockerfile ensure that nginx stay in the foreground, otherwise container will stop immediately after starting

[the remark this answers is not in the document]
bartoszbetka: Ok, I see that it's default option in openresty:alpine-fat image, so We don't need it, in our Dockerfile.

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: I suppose that, when `$backend_url` variable was set in nginx config, request `content-type` header send by user change location to request body and request `content-type` header is set to `application/x-www-form-urlencoded`. You can see this https://github.com/pgaertig/nginx-big-upload/blob/master/lib/backend_file_storage_handler.lua

cameel: This is not it.

If you look closely at the code you can see that `end_backend()` is called from within your `/upload-internal/` section, just when nginx finishes receiving the file from the client and before it starts sending it to the backend. The code you're pointing at sets the `Content-Type` header that **goes to the backend** - i.e. the `/result` section. The original `Content-Type` header gets overwritten. 

Before that (i.e. in `/upload/`) `nginx-big-upload` does not even look at this header.

So it's like I said. `/upload/ ignores `Content-Type`. Please update your description because it's misleading.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: OK so `internal` won't work. But we really can't leave it publicly accessible. Otherwise someone could bypass `/upload/` and send files directly to `/upload-internal/`, possibly using his own value for `X-Session-Id` or skipping gatekeeper checks.

Please try replacing `/upload-internal/` with `@upload-internal`. If that does not work, we'll need to find some other way to make it inaccessible.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,94 @@
...
+        set $bu_sha1     off;
+
+        set $file_storage_path /srv/storage;
+        set $backend_url       /result;
+        set $package_path      '/opt/nginx-big-upload/lib/?.lua';
+        content_by_lua_file    /opt/nginx-big-upload/lib/big-upload.lua;
+
+    }
+
+    location /result {
```
cameel: The trailing `/` is missing here. Should be `/result/` for consistency with other endpoints.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: You changed all three cases to 415. This is incorrect. Only the first one check media type. The other two should be HTTP 400.

cameel: Looks like both set the same thing but there's a big caveat: if you run `ngx.say()` nginx immediately sends response headers to the client. This means that `ngx.exit()` can no longer change the status. Nginx will print a warning if you give it a code different than the one you used in `ngx.status`:

```
[error] 12251#0: *1 attempt to set status 400 via ngx.exit after
```

See https://github.com/openresty/lua-resty-redis/issues/15#issuecomment-14747652

cameel: One more thing. The recommented style is to use `return` with `ngx.exit()`:

``` lua
return ngx.exit(400)
```

From [docs for ngx.exit()](https://github.com/openresty/lua-nginx-module#ngxexit):
> Also note that this method call terminates the processing of the current request and that it is recommended that a coding style that combines this method call with the return statement, i.e., return ngx.exit(...) be used to reinforce the fact that the request processing is being terminated.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -0,0 +1,94 @@
...
+    client_max_body_size     {{ max_upload_size }};
+
+
+    location = /upload/ {
+        access_by_lua_block {
+            -- We're using nginx-big-upload module which supports only "application/x-www-form-urlencoded".
+            -- Return an error if the content type is different.
+            if ngx.req.get_headers()["Content-Type"] ~= "application/x-www-form-urlencoded" then
+                ngx.status = 415
+                ngx.say('Unsupported content type. The only supported type is application/x-www-form-urlencoded.')
```
cameel: You're returning a text string here. It should be a JSON message, with appropriate `Content-Type`.

kubernetes/services/nginx-storage.yml.j2
```diff
@@ -38,6 +38,12 @@ spec:
         - name: nginx-storage-config
           configMap:
             name: nginx-storage-config
+      # securityContext allows you to add permission for pod and mounted volumes to specific group.
+      # We need this permission to create files and directories on mounted persistent disk.
+      # Group "65534" is the default group for user nobody in nginx-storage docker.
+      # User nobody is responsible for creating files and directories.
+      securityContext:
+        fsGroup: 65534
```
cameel: How do you know it's 65534 and what makes you think that this ID will not change? Is the value guaranteed?

cameel:
I have squashed all the fixups until now. Here's a summary of the remaining issues:

### Security
- [Publicly available /upload-internal/](https://github.com/golemfactory/concent-deployment/pull/24#discussion_r159234342)
        
### Implementation issues
- [Why 65534 as user ID](https://github.com/golemfactory/concent-deployment/pull/24/files#r159248663)

### API
- [Errors are not returned as JSON](https://github.com/golemfactory/concent-deployment/pull/24/files#r159245627)
- [Missing trailing `/` in the `/result` endpoint](https://github.com/golemfactory/concent-deployment/pull/24/files#r159234550)
- [Incorrect HTTP 415 for non-media-type-checks](https://github.com/golemfactory/concent-deployment/pull/24#discussion_r159237114)

### Optimizations
- [Clear OPM and LuaRocks cache](https://github.com/golemfactory/concent-deployment/pull/24#discussion_r157483568)

### Style
- [Misleading comment about `application/x-www-form-urlencoded`](https://github.com/golemfactory/concent-deployment/pull/24#discussion_r159199534)
- [Use `return` with `ngx.exit()`](https://github.com/golemfactory/concent-deployment/pull/24#discussion_r159242933)

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/services/nginx-storage.yml.j2
bartoszbetka: This `65534` ID is commonly reserved for user `nobody` on mostly linux version. Lot of sources confirm that, example:
- https://www.debian.org/doc/debian-policy/#uid-and-gid-classes

- http://www.linfo.org/uid.html

- https://en.wikipedia.org/wiki/Group_identifier

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: Ok. So, it's look like solution work, but require change for stripping `X-Session-Id`.

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2

containers/nginx-storage/install-dependencies.sh
```diff
@@ -0,0 +1,22 @@
...
+# but in alpine version the ldconfig isn't work properly, so we make it manually.
+ln -s /usr/lib/libcrypto.so.41 /usr/lib/libcrypto.so
+
+# Github repository with nginx-big-upload module
+git clone https://github.com/pgaertig/nginx-big-upload.git /opt/nginx-big-upload
+
+#Lua libs focusing on input data handling, functional programming and OS path managment
+/usr/local/openresty/luajit/bin/luarocks install penlight
+
+rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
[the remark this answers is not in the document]
bartoszbetka: I don't see any solution. I find only command `opm clean dist` but it's connect with `opm build`.

kubernetes/services/nginx-storage.yml.j2
cameel: OK. Nice. You should have written it in the comment in the code.

cameel: I have added it in a comment.

Issue #25, opened 2017-12-15
Script for testing file transfers on the cluster


cameel:
Create a simple script for testing file uploads and downloads on a live storage cluster.

The script should:
- Upload multiple files to `nginx-storage` and download them when the upload is complete.
- The script should accept the following parameters:
    - Server address
    - File size (default: 1 GB)
    - The number of files (default: 5)
- Generate the files on the fly, without storing them on disk. Use a command that will give you the same output every time so that it's easy to verify. For example:
    ``` bash
    seq 1000000000000000000000000000000000000000000000000000 | head --bytes 1000000000
    ```
- Compare the received data with what you sent, also without storing it on disk if possible.
- Print the time of each upload and download and of the whole script execution (see the code snippet below).
- For now ignore tokens. This script will need to be extended when we add tokens. This will be a separate task.

### Measuring execution time in bash
``` bash
start_time="$(date +%s)"
sleep 5
end_time="$(date +%s)"
echo $(("$end_time" - "$start_time"))
```

cameel:
Done in #27 and merged into master.

PR #27, opened 2017-12-20
 Script for testing file transfers on the cluster

bartoszbetka:
#25

### DIFF
--- a/kubernetes/testing-scripts/test_file_transfers.sh
+++ b/kubernetes/testing-scripts/test_file_transfers.sh
@@ -1,0 +1,69 @@
+#! /bin/bash -e
+
+DEFAULT_FILE_SIZE=1000000000
+DEFAULT_FILE_COUNT=5
+
+server_address="$1"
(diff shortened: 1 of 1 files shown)


kubernetes/testing-scripts/test_file_transfers.sh
```diff
@@ -0,0 +1,61 @@
...
+while ((iterator<="$number_of_files")); do
+    start_upload_time="$(date +%s)"
+    file_generator "$size_of_file" |                                \
+        curl                                                        \
+            --request POST                                          \
+            --connect-timeout 10                                    \
+            --include                                               \
+            --header "Concent-Upload-Path: test-file-"$iterator""   \
+            --data-binary @-                                        \
+            "http://"$server_address"/upload/"
```
cameel: This should be `"http://$server_address/upload/"`

Bash will evaluate a variable inside the double quotes just fine. And leaving it outside the quotes defeates the purpose since it won't let you use spaces inside the variable.

PR #30, opened 2017-12-21
Downloading files from nginx-storage without authentication

Gr33ny:


kubernetes/config-maps/nginx-storage/default.conf
```diff
@@ -7,7 +7,12 @@ server {
     gzip_types      text/plain text/css text/javascript application/xml;
 
 
-    location ~ ^/(download|upload)/ {
+    location ~ ^/download)/ {
```
cameel: Why is there a stray `)` in the path?

Gr33ny: Repaired.

Issue #12, opened 2017-11-28
Use Docker's cache more effectively to prevent constantly redownloading Python dependencies of concent-api in development


cameel:
Currently any change in the code of the API server (https://github.com/golemfactory/concent/) triggers redownload and installation of all Python dependencies even if the requirements did not change. For now it's still manageable but as we add dependencies it will start taking more and more time to finish. When deploying we want to rebuild from scratch anyway but for development we could avoid wasting time and bandwidth this by using Docker's cache in a smarter way.

- In `containers/Makefile` take `requirements.lock` that normally gets into `concent-api.tar` and place it in `build/concent-api/` with source files for the container.
- In `Dockerfile` for `concent-api` add a `COPY` instruction that adds only the requirements to the container.
- Run `install-concent-api.sh` before `ADD concent-api.tar`.

This way Docker will not rebuild the virtualenv unless the content of the requirements file changes.

cameel:
Merged into master: 4cae19bc2549ab6a8a9e2796603e9e7ffce561ad

PR #32, opened 2018-01-05
Improve docker caching for concent api

Gr33ny:
https://github.com/golemfactory/concent-deployment/issues/12


containers/Makefile
```diff
@@ -59,6 +59,8 @@ concent-api-clean:
 build/concent-api/concent-api.tar: $(CONCENT_DIR)
 	mkdir --parents $(dir $@)
 	package-builder/build-concent-api.sh "$(CONCENT_DIR)" "build/concent-api"
+	tar --extract --file build/concent-api/concent-api.tar \
+	--directory build/concent-api/ concent_api/requirements.lock --strip=1
```
cameel: Creating the `requirements.lock` file must be a separate target. Otherwise if the command fails make will not notice it.

cameel: Creating the `requirements.lock` file must be a separate target. Otherwise if the command fails make will not notice it.

Gr33ny: Done! Working without any problems.

PR #36, opened 2018-01-08
Add statement which check if file exist using HEAD method

### LINKED ISSUE #34: HEAD requests in nginx-storage
Concent API needs a way to check if all necessary files have been uploaded to the storage cluster. Currently it's not possible without attempting to download each file. We're going to make it possible with [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) requests. Concent API will issue a download a token for itself and ask `nginx-strorage` about each file through the public `/download/` endpoint.

- `nginx-storage` should accept HEAD requests to `/download/` endpoint and handle them the same way it does for GET requests. Use gatekeeper for authentication, check existence of the file, return 200, 404 or a different HTTP status depending on whether the file exists and if there was an error or not.
    - Remember that all error responses should be in JSON format.
    - Make sure that the request is handled correctly - the body of the response must not contain the file.
- `gatekeeper` app (https://github.com/golemfactory/concent/issues/6) should be able to handle HEAD requests to `/download/` the same way it does for GET requests.
- Extend the script for testing file transfers (#25) to make a HEAD request before downloading each file and check the HTTP status.

bartoszbetka:
Resolves #34

### DIFF
--- a/kubernetes/testing-scripts/test_file_transfers.sh
+++ b/kubernetes/testing-scripts/test_file_transfers.sh
@@ -42,6 +42,19 @@ printf "\n\n%s\n\n\n\n" "Total upload time: ${total_upload_time}s"
 i=1
 total_download_time=0
 while ((i <= "$file_count")); do
+    http_status="$(
+        curl                                                     \
+            --head                                               \
+            --silent                                             \
+            --output    /dev/null                                \
+            --write-out "%{http_code}\n"                         \
+            "http://$server_address/download/a/b/c/test-file-$i" \
+    )"
+    if [[ "$http_status" == "404" ]]; then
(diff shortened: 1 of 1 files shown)


kubernetes/testing-scripts/test_file_transfers.sh
```diff
@@ -42,6 +42,20 @@ printf "\n\n%s\n\n\n\n" "Total upload time: ${total_upload_time}s"
 i=1
 total_download_time=0
 while ((i <= "$file_count")); do
+    http_status="$(
+                    curl                                                                            \
+                        -I                                                                          \
+                        --head                                                                      \
```
cameel: `--head` and `-I` are the same thing.

PR #51, opened 2018-01-26
Geth liveness and readiness probe

### LINKED ISSUE #44: Liveness and readiness probes for geth
See [Configure Liveness and Readiness Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) in Kubernetes docs.

1) Create a liveness probe that checks if geth is running and responding to requests. If it's not, Kubernetes will assume that it hung and restart it.

2) Create a readiness probe that checks if geth is ready to accept transactions and return information over RPC. It should take into account the state of blockchain synchronization.

To implement 2) you need to determine when the node is ready. Is it after a predefined number of blocks? When the node is fully synced? Right after the node starts? Find that out.

Then use one of the APIs to get the information you need. Try the following:
- See [go-ethereum > Management APIs > admin_nodeinfo](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#admin_nodeinfo)
- See [ethereum > JSON-RPC > eth_syncing](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_syncing)
- See [How to check my sync status on Geth 1.5.5, Raspberry Pi 3](https://ethereum.stackexchange.com/questions/11233/how-to-check-my-sync-status-on-geth-1-5-5-raspberry-pi-3)

bartoszbetka:
Resolves #44


containers/geth/syncing-blockchain.sh
```diff
@@ -0,0 +1,19 @@
+#! /bin/sh -e
+
+response="$(
+    wget                                                                            \
+        -O-                                                                         \
+        --quiet                                                                     \
+        --post-data='{"jsonrpc":"2.0","method":"eth_syncing","params":[],"id":4}'   \
+        --header="Content-Type:application/json"                                    \
+        "geth.default.svc.cluster.local:8545"                                                            \
```
cameel: Could you make this address a command-line argument?

containers/geth/syncing-blockchain.sh
```diff
@@ -0,0 +1,19 @@
...
+        | jq '.result'                                                              \
+)"
+
+if [[ "$response" == "false" ]]; then
+    echo READY
+    exit 0
+fi
```
cameel: What if `$response` is something else than `true` or `false`?

Currently the script will treat is as `true` which is not a good idea in my opinion. If we get something unexpected, it's likely to be an error. Maybe the best option would be to return 2 and print an error message to stderr, saying that the response was unexpected.

containers/geth/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ethereum/client-go:{{ geth_version }}
+
+COPY syncing-blockchain.sh /usr/local/bin/
```
cameel: `is-geth-synchronized.sh` would be a better name for the script.

containers/geth/syncing-blockchain.sh
```diff
@@ -0,0 +1,19 @@
...
+        -O-                                                                         \
+        --quiet                                                                     \
+        --post-data='{"jsonrpc":"2.0","method":"eth_syncing","params":[],"id":4}'   \
+        --header="Content-Type:application/json"                                    \
+        "geth.default.svc.cluster.local:8545"                                                            \
+        | jq '.result'                                                              \
+)"
+
+if [[ "$response" == "false" ]]; then
+    echo READY
```
cameel: Maybe `SYNCHRONIZED`/`NOT SYNCHRONIZED` would be better here?

kubernetes/create-services.sh.j2
```diff
@@ -34,3 +33,4 @@ kubectl create --record --filename services/nginx-proxy.yml
 ./wait-until-ready.sh conductor-worker     30
 ./wait-until-ready.sh nginx-storage        30
 ./wait-until-ready.sh nginx-proxy          30
+./wait-until-ready.sh geth                 60
```
cameel: If this is going to wait for geth to sync then it might not be the best idea. Since it often takes more than 60 seconds, this will cause a lot of failed deployments.

kubernetes/services/geth.yml.j2
```diff
@@ -30,6 +30,20 @@ spec:
...
+              httpHeaders:
+                - name:  Content-Type
+                  value: application/json
+            initialDelaySeconds: 5
+            periodSeconds:       8
+          readinessProbe:
+            exec:
+              command: ["/usr/local/bin/syncing-blockchain.sh"]
+            initialDelaySeconds: 70
+            periodSeconds:       10
```
cameel: Isn't ` initialDelaySeconds: 70` a little pessimistic? What if Geth is up before that time?

kubernetes/create-services.sh.j2
bartoszbetka: What do you suggest in this situation ?

kubernetes/services/geth.yml.j2
bartoszbetka: This is average time of geth is up and start a synchronization. The problem is that we don't wait for geth running, but for synchronization start up.

containers/geth/syncing-blockchain.sh
```diff
@@ -0,0 +1,19 @@
+#! /bin/sh -e
+
+response="$(
+    wget                                                                            \
+        -O-                                                                         \
+        --quiet                                                                     \
+        --post-data='{"jsonrpc":"2.0","method":"eth_syncing","params":[],"id":4}'   \
```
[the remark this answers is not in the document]
bartoszbetka: It's seems that bourne shell not accept this solution, I modified it.

kubernetes/services/geth.yml.j2
cameel: It would be best to include this information in a comment.

containers/geth/is-geth-synchronized.sh
```diff
@@ -0,0 +1,36 @@
+#! /bin/sh -e
+geth_address="$1"
+
+
+message="$(cat<<EOF
+{
+    "jsonrpc": "2.0",
+    "method":  "eth_syncing",
+    "params":  [],
```
cameel: You could indent the `{ }` block for readablity. The extra whitespace should not break anything.

containers/geth/is-geth-synchronized.sh
```diff
@@ -0,0 +1,36 @@
...
+    wget                                                                            \
+        --output-document -                                                         \
+        --quiet                                                                     \
+        --header    "Content-Type:application/json"                                 \
+        --post-data "$message"                                                      \
+        "$geth_address"                                                             \
+)"
+
+synchronization_status="$(echo $response | jq '.result')"
+if [[ "$synchronization_status" == null ]]; then
```
cameel: What if it's something else than `null`? In bash `null` is just a string. You could get some other string here. I mean, you probably shouldn't get anything other than `true`/`false` if the command is correct but we're talking about handling unexpected stuff after all.

I think the condition should look like this:

```
If true
    ...
if false
    ...
else
    ...
```

containers/geth/is-geth-synchronized.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo "The response from geth failed with an error: \"$response\""
```
cameel: Please reword:
```
Invalid response from geth's 'eth_syncing' endpoint: ...
```

containers/versions.yml
```diff
@@ -3,7 +3,7 @@ docker_version:               "17.05.0~ce-0~debian-stretch"
 geth_version:                 "v1.8.12"
 golem_version:                "0.18.2"
 # Remember set version in concent_version variable also require configure it in the var-concent-<cluster_name>.yml file
-concent_version:              "v0.10.2-9-gccfae9af"
+concent_version:              "v0.10.2-71-gbcb9fdc6"
```
cameel: This looks like something from some other pull request.

containers/geth/is-geth-synchronized.sh
```diff
@@ -0,0 +1,36 @@
...
+    exit 2
+fi
+if [[ "$synchronization_status" == "false" ]]; then
+    echo SYNCHRONIZED
+    exit 0
```
cameel: Wait, does `false` really mean "synchronized" here?

kubernetes/create-services.sh.j2
bartoszbetka: The problem is when deploying process is end, when the all pods are on the cluster or when the all pods are ready to communicate with each other without error.

On which level we want to catch the problem with geth. Suspend deployment to wait for geth is ready to go or some concent service which is check connection to geth.

containers/geth/is-geth-synchronized.sh
bartoszbetka: In this case they are only three possibly value: true, false, null. The value is return by `jq` that parse specific json key. The bash only handle a result. Are you sure we want change the solution ?

containers/versions.yml
bartoszbetka: This is a test stuff.

kubernetes/create-services.sh.j2
cameel: OK. It's not a problem to solve now. Let's leave it as is and we'll worry about it when we run into problems with it.

containers/geth/is-geth-synchronized.sh
cameel: OK. I see the problem. There's no `true`. It returns a dict. In that case check for `false`, `null` and everything else.

PR #60, opened 2018-02-02
Validating hashes of uploaded file

### LINKED ISSUE #54: Checking checksums and sizes of uploaded files
`nginx-storage` should compute checksums of uploaded files while they're being uploaded. When the upload finishes, it should compare the checksum and file size with the ones stated in the token. If they don't match, reject the upload and return an error.

- `nginx-storage` doesn't have the means to decode golem messages. To work around it we'll have gatekeeper read the values from the token and return them in HTTP headers on the response (https://github.com/golemfactory/concent/issues/77). `nginx-storage` should read those values and use them in the comparison.
    - The headers are callled `Concent-File-Checksum` and `Concent-File-Size`.
    - The server should not read cheksum or file size from any headers on the request sent by the client. Only the ones from gatekeeper's response should be taken into account.
- The file should be moved to its final location only if both the checksum and file size match.
- If either the checksum or file size do not match, the file should be immediately deleted. The client should get a HTTP 400 error with a JSON message explaining why the file was rejected.
- No matter whether the operation succeeds or not, there should not be any leftover files after the operation (like for example temporary checksum files).
- Checksum must be computed incrementally, during upload, not in one operation afterwards. Otherwise we would block processing of other requests by the server.
- Checksum must begin with the name of the algorithm followed by a colon. e.g. `sha1:2bebc22296f03225617704d29d277c9e96fafcc2`. Checksums without algorithm name should not be accepted. The only algorithm supported right now is `sha1`. We'll add others later if they're needed.

bartoszbetka:
Resolves #54

### DIFF
--- a/kubernetes/config-maps/nginx-storage/default.conf.j2
+++ b/kubernetes/config-maps/nginx-storage/default.conf.j2
@@ -31,15 +31,27 @@ server {
 
     location = /upload/ {
         auth_request /upload-auth/;
+        auth_request_set $checksum $upstream_http_concent_file_checksum;
         error_page 401 = @error_401;
 
         access_by_lua_block {
+            -- Add headers from gatekeeper response
+               ngx.req.set_header("Concent-Checksum", ngx.var.checksum)
+
             -- Make sure that Concent-Upload-Path header with path and filename were included with the request.
             if ngx.req.get_headers()["Concent-Upload-Path"] == nil then
                 ngx.status = 400
                 ngx.header.content_type = "application/json; charset=utf-8"
                 ngx.say('{"error_message": "Concent-Upload-Path header is missing.", "error_code": "missing-header"}')
                 return ngx.exit(400)
+            end
+
+            -- Make sure that Concent-Checksum header was included with the request.
+            if ngx.req.get_headers()["Concent-Checksum"] == nil then
+                ngx.status = 502
+                ngx.header.content_type = "application/json; charset=utf-8"
+                ngx.say('{"error_message": "Concent-Checksum header is missing.", "error_code": "missing-header"}')
+                return ngx.exit(502)
             end
 
             -- Make sure that the request actually has a body.
@@ -51,9 +63,15 @@ server {
             end
 
             -- Strip the X-Session-Id header from the request if client has provided it.
-            -- nginx-big-upload uses his header to set the name of the temporary file.
+            -- nginx-big-upload uses this header to set the name of the temporary file.
             -- If it's empty, a random name will be generated and that's exactly what we want.
             ngx.req.set_header("X-Session-Id", "")
+
+            -- Strip the X-SHA1 header from the request if client has provided it.
+            -- nginx-big-upload uses this header to compare hash provided by client to the hash of the uploaded file.
+            -- In our case the hash is not supposed to be provided by the client so we disable this by removing the header even if the client sends it.
+            -- We get it from gatekeeper and letting client set it would allow it to cheat.
+            ngx.req.set_header("X-SHA1", "")
         }
 
         try_files $uri @upload-internal;
@@ -71,7 +89,7 @@ server {
         lua_code_cache   on;
         set $storage     backend_file;
         set $bu_checksum on;
-        set $bu_sha1     off;
+        set $bu_sha1     on;
 
         set $file_storage_path /srv/storage;
         set $backend_url       /result/;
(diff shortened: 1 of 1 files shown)


kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -42,6 +48,22 @@ server {
                 return ngx.exit(400)
             end
 
+            -- Make sure that Concent-Checksum header was included with the request.
+            if ngx.req.get_headers()["Concent-Checksum"] == nil then
+                ngx.status = 400
+                ngx.header.content_type = "application/json; charset=utf-8"
+                ngx.say('{"error_message": "Concent-Checksum header is missing.", "error_code": "missing-header"}')
```
cameel: This should be HTTP 502. The error is not on client's side but on ours.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -42,6 +48,22 @@ server {
...
+                return ngx.exit(400)
+            end
+
+            -- Make sure that Concent-File-Size header was included with the request.
+            if ngx.req.get_headers()["Concent-File-Size"] == nil then
+                ngx.status = 400
+                ngx.header.content_type = "application/json; charset=utf-8"
+                ngx.say('{"error_message": "Concent-File-Size header is missing.", "error_code": "missing-header"}')
```
cameel: Same here. HTTP 502.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -95,8 +122,42 @@ server {
             local utils = require('pl.utils')
             local table_utils   = require('table_utils')
 
-            data          = ngx.req.get_body_data()
-            tmp_file_name = string.match(data, 'id=(.*)&')
+            body_data                       = ngx.req.get_body_data()
+            tmp_file_name                   = string.match(body_data,'id=(.*)&')
+            sha1_uploaded_file              = string.match(body_data,'sha1=(.-)&')
```
cameel: Fix your style here.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -95,8 +122,42 @@ server {
             local utils = require('pl.utils')
             local table_utils   = require('table_utils')
 
-            data          = ngx.req.get_body_data()
-            tmp_file_name = string.match(data, 'id=(.*)&')
+            body_data                       = ngx.req.get_body_data()
```
cameel: Call this `nginx_big_upload_request_body`.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: This variable is not really needed.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -95,8 +122,42 @@ server {
...
+            tmp_file_name                   = string.match(body_data,'id=(.*)&')
+            sha1_uploaded_file              = string.match(body_data,'sha1=(.-)&')
+            sha1_uploaded_file_with_prefix  = "sha1:" .. sha1_uploaded_file
+            size_uploaded_file              = string.match(body_data,'size=(.-)&')
```
cameel: Call this `file_size_from_nginx_big_upload`

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -95,8 +122,42 @@ server {
...
+
+            -- Now that the file has successfully finished uploading, we can check if the hashes of file are the same.
+            -- Accept the name from Concent-Upload-Path header. We assume that this header was validated by
+            -- the gatekeeper app and is safe to use.
+            client_sha1 = ngx.req.get_headers()["Concent-Checksum"]
```
cameel: Call this `checksum_from_gatekeeper`.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -95,8 +122,42 @@ server {
...
+                os.remove("/srv/storage/" .. tmp_file_name)
+                os.remove("/srv/storage/" .. tmp_file_name .. ".shactx")
+                return ngx.exit(400)
+           end
+
+
+           -- Now that the file has successfully finished uploading, we can check if the size of file are the same.
+           -- Accept the name from Concent-File-Size header. We assume that this header was validated by
+           -- the gatekeeper app and is safe to use.
+           client_size= ngx.req.get_headers()["Concent-File-Size"]
```
cameel: Call this `file_size_from_gatekeeper`.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -95,8 +122,42 @@ server {
...
+           if  size_uploaded_file ~= client_size then
+                ngx.status = 400
+                ngx.header.content_type    = "application/json; charset=utf-8"
+                ngx.header["X-SHA1"]       = " "
+                ngx.header["X-Session-Id"] = " "
```
cameel: Why a single space? Can't it be set to an empty string?

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -95,8 +122,42 @@ server {
             local utils = require('pl.utils')
             local table_utils   = require('table_utils')
 
-            data          = ngx.req.get_body_data()
-            tmp_file_name = string.match(data, 'id=(.*)&')
+            body_data                       = ngx.req.get_body_data()
+            tmp_file_name                   = string.match(body_data,'id=(.*)&')
+            sha1_uploaded_file              = string.match(body_data,'sha1=(.-)&')
+            sha1_uploaded_file_with_prefix  = "sha1:" .. sha1_uploaded_file
```
cameel: Call this `checksum_from_nginx_big_upload`.

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: With empty string not working, default header is return in response.

cameel: Then please add a comment explaining it. It's not obvious.

cameel:
Please rebase the branch on master and squash the fixups.

PR #62, opened 2018-02-09
Ssl for nginx proxy nginx storage

### LINKED ISSUE #55: SSL for nginx-proxy and nginx-storage
Configure `nginx-proxy` and `nginx-storage` to use SSL certificates provided to them via Kubernetes secrets. We're going to terminate SSL at nginx containers. So the client connects to nginx over HTTPS and nginx connects to all the backend services over plain HTTP.

- Configure nginx to accept connections over HTTPS
- When HTTPS is enabled, the normal HTTP traffic should be redirected to HTTPS.
- Add settings in `var.yml` called `nginx_proxy_ssl` and `nginx_storage_ssl`. In configuration templates for nginx containers use these parameters to either enable SSL or stay at the current HTTP configuration.
    - By default `nginx_proxy_ssl` should be enabled and `nginx_storage_ssl` disabled (only the one at the entry point to the cluster should require SSL).
- For now use self-signed certificates. Certificates and private keys you generated should be provided to our nginx instances via Kubernetes secrets.
- Change `STORAGE_CLUSTER_ADDRESS` setting for `concent-api` to use `https://` protocol if SSL is enabled.
- Add instructions for generating a self-signed certificate to README.

Read:
- [NGINX SSL Termination](https://www.nginx.com/resources/admin-guide/nginx-ssl-termination/)
- [OpenSSL Essentials: Working with SSL Certificates, Private Keys and CSRs ](https://www.digitalocean.com/community/tutorials/openssl-essentials-working-with-ssl-certificates-private-keys-and-csrs)
- [How To Secure Nginx with Let's Encrypt on Ubuntu 16.04](https://www.digitalocean.com/community/tutorials/how-to-secure-nginx-with-let-s-encrypt-on-ubuntu-16-04)

bartoszbetka:
Resolves #55


kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -1,7 +1,23 @@
...
+
+server {
+    listen 443      ssl http2 default_server;
+    listen [::]:443 ssl http2 default_server;
+    include         snippets/nginx-proxy-ssl.conf;
+{% endif %}
+{% if nginx_proxy_ssl == "disable" %}
 server {
     listen 80;
+{% endif %}
```
cameel: Why no `default_server` here?

kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -1,7 +1,23 @@
...
+    listen [::]:80 default_server;
+    server_name    {{ cluster_address }};
+    return 302     https://$server_name$request_uri;
+
+}
+
+server {
+    listen 443      ssl http2 default_server;
+    listen [::]:443 ssl http2 default_server;
+    include         snippets/nginx-proxy-ssl.conf;
```
cameel: This is already inside the `nginx-proxy` subdirectory (in the source) or in the `nginx-proxy` image (when built by docker) so you can call this just `ssl.conf`.

kubernetes/Makefile
```diff
@@ -1,26 +1,32 @@
...
+    build/config-maps/nginx-proxy/nginx-proxy-ssl.conf      \
+    build/config-maps/gatekeeper/local_settings.py          \
+    build/config-maps/concent-api/local_settings.py         \
+    build/config-maps/nginx-storage/default.conf            \
+    build/config-maps/nginx-storage/nginx-storage-ssl.conf  \
+    build/concent-secrets/secrets.py                        \
+    build/concent-secrets/nginx-proxy-ssl.crt               \
+    build/concent-secrets/nginx-proxy-ssl.key               \
+    build/concent-secrets/nginx-storage-ssl.crt             \
+    build/concent-secrets/nginx-storage-ssl.key             \
```
cameel: Certificates and keys should not be required if SSL is disabled.

kubernetes/create-config-maps.sh
```diff
@@ -1,9 +1,27 @@
 #!/bin/bash -e
 
 
+kubectl create configmap nginx-storage-snippets                                           \
+    --from-file=nginx-storage-ssl.conf=config-maps/nginx-storage/nginx-storage-ssl.conf
```
cameel: You should put it in the same config map with the main config.

kubernetes/create-config-maps.sh
```diff
@@ -1,9 +1,27 @@
...
+
+kubectl create secret generic nginx-storage-cert                              \
+    --from-file=nginx-storage-ssl.crt=concent-secrets/nginx-storage-ssl.crt
+
+kubectl create secret generic nginx-storage-key                               \
+    --from-file=nginx-storage-ssl.key=concent-secrets/nginx-storage-ssl.key
```
cameel: If possible, put those two files in a single secret.

kubernetes/services/nginx-storage.yml.j2
```diff
@@ -30,7 +30,27 @@ spec:
               name:       concent-storage
             - mountPath:  /etc/nginx/conf.d/
               name:       nginx-storage-config
+            - mountPath:  /usr/local/openresty/nginx/conf/snippets/
+              name:       nginx-storage-snippets
+              readOnly:   true
```
cameel: Why is this config file not in `/etc/nginx/` along with the rest of nginx config?

cameel:
One more thing. Try to extract repeating code into a snippets that can be shared between both servers. We don't want repeated code.

kubernetes/services/nginx-storage.yml.j2
bartoszbetka: Nginx in openresty docker is in a different location ` /usr/local/openresty/nginx/`. However  `nginx.conf` is set to add configs from `/etc/nginx/conf.d` directory. I move config to `/etc/nginx/`.

kubernetes/config-maps/gatekeeper/local_settings.py.j2
```diff
@@ -35,4 +35,10 @@ SERVER_EMAIL         = "{{ email.server_email }}"
 
 CONCENT_FEATURES = ["gatekeeper"]
 
+{% if nginx_storage_ssl == "disabled" %}
 STORAGE_CLUSTER_ADDRESS = "http://{{ storage_cluster_address }}/"
+{% endif %}
+
+{% if nginx_storage_ssl == "enabled" %}
+STORAGE_CLUSTER_ADDRESS = "https://{{ storage_cluster_address }}/"
+{% endif %}
```
cameel: Use `{% else %}`. `nginx_storage_ssl` should be a boolean value.

kubernetes/services/nginx-storage.yml.j2
```diff
@@ -29,15 +29,27 @@ spec:
             - mountPath:  /srv/storage/
               name:       concent-storage
             - mountPath:  /etc/nginx/conf.d/
-              name:       nginx-storage-config
+              name:       nginx-storage-configs
+              readOnly:   true
```
cameel: You can't mount the whole map inside `conf.d/` because it contains `ssl.conf` which is not a standalone config. If you do this, nginx will always include `ssl.conf` at the top level. Put `ssl.conf` in a subdirectory (and make sure nginx does not load configs from subdirectories).

cameel: OK. After talking with @bartoszbetka we're going to leave it as is. It's not included by the main config any more.

bartoszbetka: Adding `ssl.conf` in `conf.d` folder causes that all configs in `http` directive depend on it. We use only one config with `server` directive, so there is no problem, to keep `ssl.conf` in `conf.d/` folder.

PR #67, opened 2018-02-19
Update cluster to use python 3.6 in concent-api

### LINKED ISSUE #31: Python 3.6 on the cluster
Currently we're using Debian Stretch which offers only Python 3.5. We want to use 3.6. Best if we don't have to build it from source during deployment.

bartoszbetka:
Resolves #31

### DIFF
--- a/concent-builder/configure.yml
+++ b/concent-builder/configure.yml
@@ -14,6 +14,16 @@
             - ca-certificates
             - curl
 
+        - name:    Use pinning to ensure that packages from testing are installed only of stable packages are not available
+          copy:
+            src:  files/testing.pref
+            dest: /etc/apt/preferences.d/testing.pref
+
+        - name:    Enable the testing repository
+          apt_repository:
+            repo:  "deb http://ftp.de.debian.org/debian testing main"
+            state: present
+
         - name:    Add Google SDK repository key
           apt_key:
             url:   https://packages.cloud.google.com/apt/doc/apt-key.gpg
@@ -38,7 +48,7 @@
           apt:  update_cache=yes
 
         - name:    Install basic packages
-          apt:  name={{ item }}
+          apt:     name={{ item }}
           with_items:
             - ncdu
             - tree
@@ -46,29 +56,35 @@
             - tmux
 
         - name:    Install dependencies
-          apt:  name={{ item }}
+          apt:     name={{ item }}
           with_items:
             - build-essential
             - python2.7
-            - python3
-            - python-pip
+            - python-pip  # For Python 2
             - libssl-dev
+            - pkg-config
+            - libsecp256k1-dev
             - git
             - kubectl
             - docker-engine
             - google-cloud-sdk
             - ansible
-            - virtualenv
             - libpq-dev
             - gcc
-            - python3-dev
             - python3-yaml
+
+        - name:    Install dependencies from testing
+          apt:  name={{ item }} default_release=testing
+          with_items:
+            - python3.6
+            - python3.6-dev
+            - virtualenv
 
         - name:     apt autoclean
           command:  apt autoclean
 
         - name:     apt autoremove
-          command:  apt autoremove
+          command:  apt autoremove --assume-yes
 
         - name:    Install yasha in global python environment using pip
           pip:
--- a/containers/concent-api/install-concent-api.sh
+++ b/containers/concent-api/install-concent-api.sh
@@ -1,12 +1,11 @@
 #!/bin/bash -e
 
 virtualenv_dir=/srv/http/virtualenv
-python_version=3.5
+python_version=3.6
 
 # Create virtualenv and Django app's dependencies
-virtualenv --python python3 "$virtualenv_dir"
+virtualenv --python python$python_version "$virtualenv_dir"
 source "$virtualenv_dir/bin/activate"
 pip install --upgrade pip
 pip install --requirement "/srv/http/concent_api/requirements.lock"
 pip install gunicorn
-
--- a/containers/concent-api/install-dependencies.sh
+++ b/containers/concent-api/install-dependencies.sh
@@ -2,22 +2,42 @@
 
 # Update package lists for Debian and install dependencies
 dependencies=(
-    python3-pip
-    python3-virtualenv
-    virtualenv
-
     # Required by psycopg2 python package (PostgreSQL support)
     libpq-dev
 
     # Packages needed to build python extensions, dependencies of golem-messages in particular
     gcc
-    python3-dev
     libssl-dev
     git
+    pkg-config
+    libsecp256k1-dev
+    libffi-dev
+    software-properties-common
 )
-apt-get --assume-yes update
-apt-get --assume-yes install --no-install-recommends ${dependencies[*]}
+
+testing_dependencies=(
+    # There's no stable Python 3.6 package in Debian Stretch yet so we have to get it from testing.
+    python3.6
+    python3.6-dev
+    virtualenv
+)
+
+apt-get update --assume-yes
+apt-get install              \
+    --assume-yes             \
+    --no-install-recommends  \
+    ${dependencies[*]}
+
+add-apt-repository "deb http://ftp.de.debian.org/debian testing main"
+apt-get update --assume-yes
+
+apt-get install              \
+    --target-release testing \
+    --assume-yes             \
+    --no-install-recommends  \
+    ${testing_dependencies[*]}
 
 # Clean up
 apt-get clean
+apt autoremove --assume-yes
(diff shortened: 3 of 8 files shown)


containers/concent-api/install-dependencies.sh
```diff
@@ -2,22 +2,35 @@
...
+    libffi-dev
+    software-properties-common
+)
+
+# We need to use python3.6. There is no official python3.6 package on the debian stretch,
+# we must include test packages.
+python_dependencies=(
+    python3.6
+    python3.6-dev
+    python3.6-venv
```
cameel: Put this (including the comment) together with all the other dependencies.

containers/concent-api/install-dependencies.sh
```diff
@@ -2,22 +2,35 @@
...
 )
+
 apt-get --assume-yes update
 apt-get --assume-yes install --no-install-recommends ${dependencies[*]}
+add-apt-repository "deb http://ftp.de.debian.org/debian testing main"
+apt-get --assume-yes update
+apt-get --target-release testing --assume-yes install --no-install-recommends ${python_dependencies[*]}
```
cameel: Move the `add-apt-repository` line above the first `apt-get update`. This way yuo won't have to update twice.

concent-builder/configure.yml
```diff
@@ -50,25 +55,26 @@
...
             - gcc
-            - python3-dev
             - python3-yaml
 
         - name:     apt autoclean
           command:  apt autoclean
 
         - name:     apt autoremove
-          command:  apt autoremove
+          command:  apt --assume-yes autoremove
```
cameel: Always put the extra options after the command. This makes it easier to see that the command is:
``` bash
apt autoremove --assume-yes
```

containers/concent-api/install-dependencies.sh
cameel: Always put the extra options after the command:
```
apt-get update --assume-yes
apt-get install              \
    --target-release testing \
    --assume-yes             \
    --no-install-recommends  \
    ${python_dependencies[*]}
```

containers/concent-api/install-dependencies.sh
```diff
@@ -2,22 +2,35 @@
...
+# we must include test packages.
+python_dependencies=(
+    python3.6
+    python3.6-dev
+    python3.6-venv
 )
+
 apt-get --assume-yes update
 apt-get --assume-yes install --no-install-recommends ${dependencies[*]}
+add-apt-repository "deb http://ftp.de.debian.org/debian testing main"
```
cameel: Why this particular mirror? Is it the official one for testing or just one you chose?

containers/package-builder/build-concent-api-assets.sh
```diff
@@ -4,15 +4,16 @@ concent_dir=$(realpath "$1")
...
-python_version=3.5
+python_version=3.6
 
 mkdir --parents "$build_dir/concent-api-assets/"
 cp --recursive "$concent_dir/concent_api" "$build_dir/concent-api-assets"
 
 cd "$build_dir/concent-api-assets/concent_api/"
 
-virtualenv --python python3 "$build_dir/virtualenv/"
+python$python_version -m venv "$build_dir/virtualenv/"
```
cameel: Doesn't `python3-venv` provide the `virtualenv` command?

containers/package-builder/build-concent-api-assets.sh
```diff
@@ -4,15 +4,16 @@ concent_dir=$(realpath "$1")
...
 source "$build_dir/virtualenv/bin/activate"
+pip install wheel
```
cameel: Why do you need to install this explicitly?

bartoszbetka: Package `virtualenv` adds `python-setuptools-whl` lib, while `python3.6-venv` does not add it.

bartoszbetka: The `virtualenv` package adds `python-setuptools-whl` lib, while `python3.6-venv` does not add it.

containers/package-builder/build-concent-api-assets.sh
bartoszbetka: No, It does't provide `virtualenv` command.

containers/concent-api/install-dependencies.sh
bartoszbetka: This is the one of the official mirrors https://www.debian.org/mirror/list. We can also choose `ftp.pl.debian.org/debian/`.

containers/concent-api/install-dependencies.sh
bartoszbetka: I think that when we do this, all libraries will be installed from the test or default repository. The `--target-release` flag can control it, but when we add all dependencies together, we can not control it for a specific library.

containers/concent-api/install-dependencies.sh
bartoszbetka: The `software-properties-common` package allows to use `add-apt-repository`, which is in dependencies, so before we can use `add-apt-repository`, we must install depedencies.

containers/concent-api/install-dependencies.sh
```diff
@@ -2,22 +2,42 @@
...
-apt-get --assume-yes install --no-install-recommends ${dependencies[*]}
+apt-get install              \
+    --target-release testing \
+    --assume-yes             \
+    --no-install-recommends  \
+    ${python_dependencies[*]}
 
 # Clean up
 apt-get clean
+apt --assume-yes autoremove
```
cameel: Options after the command.

containers/concent-api/install-dependencies.sh
```diff
@@ -2,22 +2,42 @@
...
+
+# We need to use python3.6. There is no official python3.6 package on the debian stretch,
+# we must include test packages.
+python_dependencies=(
+    python3.6
+    python3.6-dev
+    python3.6-venv
 )
+
+apt-get --assume-yes update
```
cameel: Options after the command.

containers/concent-api/install-dependencies.sh
cameel: OK. But in that case you should make this clearer.

Rename the variable to `testing_dependencies`. Put the comment about python 3.6 inside (later we may need to use other packages from testing).

containers/concent-api/install-concent-api.sh
```diff
@@ -1,12 +1,13 @@
...
 virtualenv_dir=/srv/http/virtualenv
-python_version=3.5
+python_version=3.6
 
 # Create virtualenv and Django app's dependencies
-virtualenv --python python3 "$virtualenv_dir"
+python$python_version -m venv "$virtualenv_dir"
 source "$virtualenv_dir/bin/activate"
 pip install --upgrade pip
+pip install wheel
```
cameel: Use `python3-wheel`. Or `python3-virtualenv`.

bartoszbetka: > Currently, there are two common tools for creating Python virtual environments:
>- venv is available by default in Python 3.3 and later, and installs pip and setuptools into created virtual environments in Python 3.4 and later.
>- virtualenv needs to be installed separately, but supports Python 2.6+ and Python 3.3+, and pip, setuptools and wheel are always installed into created virtual environments by default (regardless of Python version).

https://packaging.python.org/tutorials/installing-packages/

containers/concent-api/install-dependencies.sh
```diff
@@ -2,22 +2,41 @@
...
+    virtualenv
+)
+
+apt-get --assume-yes update
+apt-get install              \
+    --assume-yes             \
+    --no-install-recommends  \
+    ${dependencies[*]}
+add-apt-repository "deb http://ftp.de.debian.org/debian testing main"
 apt-get --assume-yes update
```
cameel: This is not the best way to install stuff from testing. Any apt command after doing this will install packages from testing (even if you don't specify `--target-release testing` explicitly). This now works but it's easy for someone to come later and break it.

You should use pinning instead. Add testing repo with piority < 100. Then use `--target-release`
See https://askubuntu.com/questions/27362/how-to-only-install-updates-from-a-specific-repository/814936#814936

concent-builder/configure.yml
```diff
@@ -14,6 +14,11 @@
             - ca-certificates
             - curl
 
+        - name:    Add debian test repository for python3.6 installation
+          apt_repository:
+            repo:  "deb http://ftp.de.debian.org/debian testing main"
+            state: present
```
cameel: This permanently adds the testing repo. We don't want that. We only want to install some stuff from testing. Use the same method I suggested for the Dockerfile.

PR #69, opened 2018-02-21
Ssl optimizations and advenced configuration

### LINKED ISSUE #56: SSL optimizations and advanced configuration
After implementing #55 go to [Qualys SSL Server Test](https://www.ssllabs.com/ssltest/) and run their test against our nginx on the dev cluster. Looks at the results and list the issues in the comments below.

Before running the test I see several options that may need to be adjusted:
- Supported protocols. We should definitely disable SSL v3 and earlier. Support only TLS. You can be aggressive with your config here because we don't have to support old browsers.
- Ciphers and Diffie-Hellman parameters
- Configure session caching
- Configure OCSP stapling
- Enable [HSTS](https://developer.mozilla.org/en-US/docs/Security/HTTP_Strict_Transport_Security)

More information:
- Here's an example SSL configuration for nginx with more parameters: [nrollr > NGINX config for SSL with Let's Encrypt certs ](https://gist.github.com/nrollr/9a39bb636a820fb97eec2ed85e473d38)
- [Running Your Flask Application Over HTTPS](https://blog.miguelgrinberg.com/post/running-your-flask-application-over-https), in particular the section called "Achieving an SSL A+ Grade".

bartoszbetka:
Resolves #56


cameel:
The two `ssl.conf` files are almost identifcal. Extract repeating stuff into a single place.

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
+# Disable SSLv3
+ssl_protocols              TLSv1 TLSv1.1 TLSv1.2;
```
cameel: Please disable `TLSv1` and `TLSv1.1`.
- TLS 1.0 is vulnerable to multiple attacks and no longer considered secure. Since 2016 the recommendation is to have it disabled: [SSL and TLS 1.0 No Longer Acceptable for PCI Compliance](https://blog.varonis.com/ssl-and-tls-1-0-no-longer-acceptable-for-pci-compliance/)
- TLS 1.1 is still OK but it's never been in widely adopted (see [Deprecating TLS 1.0 & 1.1](https://www.digicert.com/blog/depreciating-tls-1-0-and-1-1/), [Is there a reason for still having TLSv1.1 enabled?](https://community.qualys.com/thread/16565-is-there-a-reason-for-still-having-tlsv11-enabled)). Any client supporting it is very likely to support TLS 1.2 too so there's no reason not to force 1.2.

And in any case, we can afford to offer only TLS 1.2 because we don't have to support ancient browsers (or any browsers at all for that matter).

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
...
+# Specify curve P-384
+ssl_ecdh_curve             secp384r1;
+
+# Improve ssl performence
+ssl_session_cache          shared:SSL:10m;
+ssl_session_timeout        5m;
+
+# Enable OCSP stapling
+ssl_stapling               on;
+ssl_stapling_verify        on;
```
cameel: Please check what happens when OCSP stapling is enabled while using a self-signed certificate. I.e. whether nginx tries to contact the stapling responder and fails.

Stapling makes sense only with certificates singed by a CA because the whole feature is just that nginx asks the CA if the certificate is still valid. If there's no CA (the certificate is self-signed), it obviously can't be done. We want it if we can have it but with self-signed certificates we can't so it's best if nginx can just ignore in that case. But if it can't, we need a variable that enables/disables stapling.

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
...
+
+# Enable HSTS
+add_header Strict-Transport-Security "max-age=63072000; includeSubdomains";
```
cameel: This `max-age` has a significant gotcha. Once we switch dev or any other cluster to HTTPS, switching back to HTTP will require clearing client's cache.

Please at least add some explanations:
```
# HSTS tells the client to only access the site over HTTPS. This is mostly relevant to browsers
# but other types clients may implement it too as long as they have some persistent storage.
# See https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet
# Note that max-age here is 63072000 seconds (2 years) which means that once you start serving
# traffic over HTTPS, you won't be able to switch back to HTTP without clearing client cache.
```

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
...
+
+# Advanced settings
+server_tokens off;
```
cameel: Add a comment here:

```
# Disable emitting nginx version on error pages and in the “Server” response header field. 
```

cameel: This is not related to SSL. It could be enabled even when SSL is not. Please move it to a separate config file (e.g. something like `global.conf`.

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
...
+add_header X-XSS-Protection "1; mode=block";
+add_header X-Frame-Options DENY;
+add_header X-Content-Type-Options nosniff;
```
cameel: These options also have nothing to do with SSL. Please move them somewhere else.

Also, please add some explanations. At least links to their definitions:

```
# Enable XSS protection for ancient browsers that don't support CSP.
# See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection
add_header X-XSS-Protection "1; mode=block";

# Prevent rendering our pages inside HTML frames.
# We don't have frames (and we don't really serve many web pages) so there's no harm in enabling it.
# See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options
add_header X-Frame-Options DENY;

# Enforce strict interpretation of `Content-Type` for files included in `<script>` and `<style>` HTML tags.
# Again, relevant only for webpages but there's no harm in having it enabled.
# See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
add_header X-Content-Type-Options nosniff;
```

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
+# Disable SSLv3
+ssl_protocols              TLSv1 TLSv1.1 TLSv1.2;
+
+# Enable server-side protection
+ssl_prefer_server_ciphers  on;
+ssl_ciphers                "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH";
```
cameel: Please add some comments explaining why you chose those specific ciphers. Best if you can add a link to some authoritative recommendation.

kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -25,6 +25,12 @@ server {
     listen 443      ssl http2 default_server;
     listen [::]:443 ssl http2 default_server;
 
+    ssl_certificate     /etc/ssl/secrets/nginx-proxy-ssl.crt;
+    ssl_certificate_key /etc/ssl/secrets/nginx-proxy-ssl.key;
+
+    # Add Diffie-Helman group
+    ssl_dhparam /etc/ssl/secrets/nginx-proxy-dhparam.pem;
```
cameel: How big is this file and how long does it take to generate it?

It would be best to generate it during container build.

kubernetes/create-config-maps.sh.j2
```diff
@@ -4,24 +4,26 @@
 kubectl create configmap nginx-config-snippets                          \
     --from-file=error-pages.conf=config-maps/nginx/error-pages.conf
 
-kubectl create secret generic nginx-storage-secrets                              \
-    --from-file=nginx-storage-ssl.crt=concent-secrets/nginx-storage-ssl.crt      \
-    --from-file=nginx-storage-ssl.key=concent-secrets/nginx-storage-ssl.key
+kubectl create secret generic nginx-storage-secrets                                 \
+    --from-file=nginx-storage-ssl.crt=concent-secrets/nginx-storage-ssl.crt         \
+    --from-file=nginx-storage-ssl.key=concent-secrets/nginx-storage-ssl.key         \
+    --from-file=nginx-storage-dhparam.pem=concent-secrets/nginx-storage-dhparam.pem
```
cameel: Is `nginx-storage-dhparam.pem` really a secret?

kubernetes/config-maps/nginx/ssl.conf
cameel: You should forbid null ciphers by adding `:!aNULL`. Maybe also forbid some very weak ciphers too.

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
...
+ssl_stapling               on;
+ssl_stapling_verify        on;
+
+# Enable HSTS
+add_header Strict-Transport-Security "max-age=63072000; includeSubdomains";
+
+# Advanced settings
+server_tokens off;
+add_header X-XSS-Protection "1; mode=block";
+add_header X-Frame-Options DENY;
```
cameel: Actually might be better to set it to `SAMEORIGIN` to allow frames in our own code while preventing embedding on other sites.

kubernetes/config-maps/nginx/ssl.conf
```diff
@@ -0,0 +1,26 @@
...
+# Enable server-side protection
+ssl_prefer_server_ciphers  on;
+ssl_ciphers                "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH";
+
+# Specify curve P-384
+ssl_ecdh_curve             secp384r1;
+
+# Improve ssl performence
+ssl_session_cache          shared:SSL:10m;
+ssl_session_timeout        5m;
```
cameel: Add a link to this site in a comment: https://vincent.bernat.im/en/blog/2011-ssl-session-reuse-rfc5077

cameel: Please disable `ssl_session_tickets` as well.

https://github.com/mozilla/server-side-tls/issues/135

kubernetes/config-maps/nginx/ssl.conf
cameel: This form does not follow the RFC. Here's an alternative:
https://gist.github.com/plentz/6737338#gistcomment-1299170

kubernetes/config-maps/nginx/ssl.conf
cameel: This generator from Mozilla can give you some good values: https://mozilla.github.io/server-side-tls/ssl-config-generator/

kubernetes/config-maps/nginx/ssl.conf
cameel: Also, keep this in mind: [Nginx add_header configuration pitfall ](https://blog.g3rt.nl/nginx-add_header-pitfall.html).

kubernetes/config-maps/nginx/ssl.conf
cameel: Also, see this: https://cipherli.st/

kubernetes/config-maps/nginx/ssl.conf
cameel: Please add a `Referrer-Policy` header too and set it to `no-referrer`.

https://scotthelme.co.uk/a-new-security-header-referrer-policy/
https://www.perpetual-beta.org/weblog/the-curious-case-of-tls-and-the-missing-referrers.html

kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -25,6 +25,12 @@ server {
     listen 443      ssl http2 default_server;
     listen [::]:443 ssl http2 default_server;
 
+    ssl_certificate     /etc/ssl/secrets/nginx-proxy-ssl.crt;
+    ssl_certificate_key /etc/ssl/secrets/nginx-proxy-ssl.key;
+
+    # Add Diffie-Helman group
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/config-maps/nginx/ssl.conf
bartoszbetka: Done.

kubernetes/config-maps/nginx/ssl.conf
bartoszbetka: This may cause a problem when traffic with HTTP and HTTPS are allowed.

kubernetes/config-maps/nginx/ssl.conf
bartoszbetka: Done.

bartoszbetka: Done.

kubernetes/config-maps/nginx/ssl.conf
bartoszbetka: Done.

kubernetes/config-maps/nginx-proxy/default.conf.j2
bartoszbetka: It's a small file which is generate fast. This solution requires additional script and causes generate new `.pem` file for new deployment.

bartoszbetka: Are you sure we want to use it ?

kubernetes/config-maps/nginx/ssl.conf
bartoszbetka: Done.

kubernetes/config-maps/nginx/ssl.conf
bartoszbetka: Done.

bartoszbetka: Done.

bartoszbetka: I use value from https://cipherli.st/

kubernetes/config-maps/nginx/ssl.conf
bartoszbetka: Ok, so I deleted this settings from config.

kubernetes/config-maps/nginx-proxy/default.conf.j2
cameel: Why not?

bartoszbetka: Why want to it? Ok, so I will do it.

bartoszbetka: I checked how much time need to generate this file and:
- `~10s for 2048bit`
- `~30s for 4096bit`

bartoszbetka: We still want this solution ?

bartoszbetka: I checked, how much time need to generate this file and it's not deterministic and random on each execution. For generate 2048bit key, it's need sometimes 10s and sometimes 40s. The openssl inform that `Generating DH parameters, 2048 bit long safe prime, generator 2
This is going to take a long time`. We still want this solution ?

PR #70, opened 2018-03-01
Add playbook and script which run django unit tests

### LINKED ISSUE #37: Running django unit tests in deployment playbooks
Create an additional Ansible playbook in `concent-builder/` that runs unit tests in `concent-api` container:
- Use `testing.py` as a base for setting (https://github.com/golemfactory/concent/issues/28).
- Run `manage.py test` in the container using `docker run`.
    - You'll have to run PostgreSQL server in another container. There are several options:
        - Run the official `postgres` image in a local container only for the duration of the test.
        - Set up a permanent PostgreSQL service on the docker builder machine
        - Use the CloudSQL server from the cluster
        - Reconfigure the application to run on SQLite instead of PostgreSQL. This may not work though since we heavily rely on transactions and rollbacks.
- Fail if not all tests pass.
- Run linter and type checks too. Fail if there are any warnings.
- Update README with instructions for running this step.

bartoszbetka:
Resolves #37


concent-builder/unit-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+              --zone    "{{ gke.zone }}"
+
+        - name:  Run postgres container for django unit tests
+          shell: docker ps -q --filter "name=postgres" | grep -q .               \
+              || docker run --rm -d --name postgres concent-postgresql:latest
+
+        - name:    Run concent-api container and django unit tests
+          command: docker run --rm                                                                                      \
+              -v {{ build_dir }}/concent-deployment/kubernetes/build/run-unit-tests.sh:/usr/local/bin/run-unit-tests.sh \
+              --link postgres concent-concent-api:latest /usr/local/bin/run-unit-tests.sh
```
cameel: `--link` option is deprecated. You should create a `docker network` instead and attach both containers to it using `--net` option.

See https://www.reddit.com/r/docker/comments/3smpwt/linking_with_link_will_be_deprecated_how_does/

kubernetes/run-unit-tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
+
+# We must change local_settings.py, because django search for configmap which is added in kubernetes.
+echo "from .testing import *" > "/srv/http/concent_api/concent_api/settings/local_settings.py"
```
cameel: Why is this necessary? Youre using `--settings=concent_api.settings.testing` below and it should be enough to make `manage.py` use `testing.py`.

kubernetes/run-unit-tests.sh
```diff
@@ -0,0 +1,8 @@
...
+# We must add host value to testing.py settings to enable communication with database.
+sed -i '/USER/a "HOST":"postgres",' /srv/http/concent_api/concent_api/settings/testing.py
```
cameel: If you need to modify values in `testing.py` you should just create another settings file that imports `testing` and define `DATABASES` setting in it.

concent-builder/unit-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+
+        - name:    Stop postgres container after tests were completed
+          command: docker stop postgres
```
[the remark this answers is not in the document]
bartoszbetka: I know, but shell command check if container exist, so that's not a problem. When playbook's fail, we won't run the next condition. I think it's not necessary to stop the postgres container when some test fails in concent-api cotainer and then again run postgres with the same configuration.

kubernetes/run-unit-tests.sh
bartoszbetka: I saw this option, but when I change `local_settings.py`, I'm sure that django use appriopriate settings.

concent-builder/unit-tests.yml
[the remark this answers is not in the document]
cameel: But if you want Postgres to run constantly then it's not the way to do this. You should just start the service in the `configure` playbook.

We want to either start and stop PostgreSQL for each test or keep it running constantly. Your solution is something in between.

kubernetes/run-unit-tests.sh
cameel: Modifying `local_settings.py` in an ad-hoc way like seems very hacky. I strongly prefer the solution with `--settings`.

concent-builder/unit-tests.yml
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/run-unit-tests.sh
cameel: You can simply mount a volume. As long as the file is there when we run the container, it's fine.

bartoszbetka: I resolved this problem by creating empty file in `.config_map.local_settings`

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+          command: gcloud container clusters get-credentials \
+              "{{ gke.cluster }}"                            \
+              --project "{{ gke.project }}"                  \
+              --zone    "{{ gke.zone }}"
+
+        - name:    Run custom network to allows communication between django and postgres containers
+          command: docker network create {{ image_prefix }}network
+
+        - name:    Run postgres container for django unit tests
+          command: docker run --rm --detach --net={{ image_prefix }}network --name {{ image_prefix }}postgresql-unittest {{ image_prefix }}postgresql:latest
```
cameel: Please wrap this command into multiple lines for readability (just like the command below).

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+
+        - name:    Stop postgres container after tests were completed
+          command: docker stop {{ image_prefix }}postgresql-unittest
+
+        - name:    Delete custom network after tests were completed
+          command: docker network rm {{ image_prefix }}network
+      rescue:
+        - name:  Stop postgres container and delete custom network after occurs error
+          shell: docker stop {{ image_prefix }}postgresql-unittest               \
+              && docker network rm {{ image_prefix }}network
```
cameel: This should be two separate steps.

kubernetes/docker_testing.py.j2
```diff
@@ -0,0 +1,15 @@
...
+        # This could happen e.g. if someone executes a query at module level.
+        'ENGINE':          'django.db.backends.postgresql_psycopg2',
+        'USER':            'postgres',
+        'HOST':            '{{ image_prefix }}postgresql-unittest',
+        'ATOMIC_REQUESTS': True,
+        'TEST': {
+            'NAME': 'test_concent_api'
+        }
+    }
+}
```
cameel: Instead of replacing the whole `DATABASES` setting, only change values in the default config.

Also, now that Conductor has been merged, please configure its database too.

kubernetes/run-unit-tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+
+mkdir "/srv/http/concent_api/concent_api/settings/config_map"
+touch "/srv/http/concent_api/concent_api/settings/config_map/local_settings.py"
+
+/usr/local/bin/concent-api-manage.sh                    \
+    "test"                                              \
+    "--failfast"                                        \
+    "--settings=concent_api.settings.docker_testing"
```
cameel: You don't really have to put these options in quotes.

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+              --volume {{ build_dir }}/concent-deployment/kubernetes/build/run-unit-tests.sh:/usr/local/bin/run-unit-tests.sh                               \
+              --volume {{ build_dir }}/concent-deployment/kubernetes/build/docker_testing.py:/srv/http/concent_api/concent_api/settings/docker_testing.py   \
+              --net={{ image_prefix }}network concent-concent-api:latest /usr/local/bin/run-unit-tests.sh
+
+        - name:    Stop postgres container after tests were completed
+          command: docker stop {{ image_prefix }}postgresql-unittest
+
+        - name:    Delete custom network after tests were completed
+          command: docker network rm {{ image_prefix }}network
+      rescue:
```
cameel: Instead of using `rescue` and repeating cleanup twice you could use `always` which is guaranteed to be always executed.

concent-builder/test.yml
cameel: Please use also `{{ cluster }}` in container name so that we can run builds in parallel for different clusters.

I.e. `{{ cluster }}-{{ image_prefix }}postgresql-unittest`

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+      become_user: "{{ shared_user }}"
+      block:
+        - name: Configure kubectl to operate on the right cluster
+          command: gcloud container clusters get-credentials \
+              "{{ gke.cluster }}"                            \
+              --project "{{ gke.project }}"                  \
+              --zone    "{{ gke.zone }}"
+
+        - name:    Run custom network to allows communication between django and postgres containers
+          command: docker network create {{ image_prefix }}network
```
cameel: Please use `{{ cluster }}` in network name.

I.e. `{{ cluster }}-{{ image_prefix }}unittest-network`

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+
+        - name:    Run postgres container for django unit tests
+          command: docker run --rm --detach --net={{ image_prefix }}network --name {{ image_prefix }}postgresql-unittest {{ image_prefix }}postgresql:latest
+
+        - name:    Run concent-api container and django unit tests
+          command: docker run --rm                                                                                                                          \
+              --volume {{ build_dir }}/concent-deployment/kubernetes/build/run-unit-tests.sh:/usr/local/bin/run-unit-tests.sh                               \
+              --volume {{ build_dir }}/concent-deployment/kubernetes/build/docker_testing.py:/srv/http/concent_api/concent_api/settings/docker_testing.py   \
```
cameel: Mount these volumes read-only. The unittests should not be able to change these files.

kubernetes/run-unit-tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+
+mkdir "/srv/http/concent_api/concent_api/settings/config_map"
+touch "/srv/http/concent_api/concent_api/settings/config_map/local_settings.py"
```
cameel: Please add a comment that explains why we need to do this:

```
# It seems that Django tries to import the default settings module even if it's told to use a different one.
# To work around it we need empty local_settings.py so that importing does not fail.
```

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+    - become:      yes
+      become_user: "{{ shared_user }}"
+      block:
+        - name: Configure kubectl to operate on the right cluster
+          command: gcloud container clusters get-credentials \
+              "{{ gke.cluster }}"                            \
+              --project "{{ gke.project }}"                  \
+              --zone    "{{ gke.zone }}"
+
+        - name:    Run custom network to allows communication between django and postgres containers
```
cameel: `Run` -> `Create`
`to allows` -> `that allows`

concent-builder/test.yml
bartoszbetka: Done.

kubernetes/docker_testing.py.j2
bartoszbetka: Done.

kubernetes/run-unit-tests.sh
bartoszbetka: Done.

concent-builder/test.yml
bartoszbetka: Done.

concent-builder/test.yml
bartoszbetka: Done.

concent-builder/test.yml
bartoszbetka: Done.

concent-builder/test.yml
bartoszbetka: Done.

kubernetes/run-unit-tests.sh
bartoszbetka: Done.

concent-builder/test.yml
bartoszbetka: Done.

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+          command: gcloud container clusters get-credentials \
+              "{{ gke.cluster }}"                            \
+              --project "{{ gke.project }}"                  \
+              --zone    "{{ gke.zone }}"
+
+        - name:    Create custom network that allows communication between django and postgres containers
+          command: docker network create {{ cluster }}-{{ image_prefix }}unittest-network
+
+        - name:    Run postgres container for django unit tests
+          command: docker run --rm                                                                               \
```
cameel: `--rm` should be at the same level as the other options.

concent-builder/test.yml
```diff
@@ -0,0 +1,37 @@
...
+              --detach                                                                                           \
+              --net={{ cluster }}-{{ image_prefix }}unittest-network                                             \
```
cameel: `--net` works just like other options. No need for `=`. And you could use the full form: `--network`

kubernetes/run-unit-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+
+# It seems that Django tries to import the default settings module even if it's told to use a different one.
+# To work around it we need empty local_settings.py so that importing does not fail.
+echo -n "" > "/srv/http/concent_api/concent_api/settings/local_settings.py"
```
cameel: After talking about it on Slack, we established that this should not be necessary. Please remove it.

concent-builder/test.yml
bartoszbetka: Done.

PR #71, opened 2018-03-02
Create and add file with concent version to container that run the Django application

### LINKED ISSUE #64: Creating a file with Concent version in all Concent containers
When https://github.com/golemfactory/concent/issues/126 is implemented, the `concent` repository will contain a script that creates a file with Concent version. This file must be present in every container that runs the Django application. Update our Dockerfiles to run the script and create the file.

bartoszbetka:
Resolves #64

### DIFF
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -68,6 +68,10 @@ postgresql-clean:
 concent-api-clean:
 	docker rmi $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) --force
 
+
+build/virtualenv/: $(CONCENT_DIR)/concent_api/requirements.lock
+	package-builder/build-virtualenv.sh "$<" "build"
+
 build/concent-api/requirements.lock: build/concent-api/concent-api.tar
 	tar                                          \
 		--extract                                \
@@ -76,12 +80,12 @@ build/concent-api/requirements.lock: build/concent-api/concent-api.tar
 		--directory build/concent-api/           \
 		concent_api/requirements.lock
 
-build/concent-api/concent-api.tar: $(CONCENT_DIR)
+build/concent-api/concent-api.tar: $(CONCENT_DIR)  build/virtualenv/
 	mkdir --parents $(dir $@)
 	package-builder/build-concent-api.sh "$(CONCENT_DIR)" "build/concent-api"
 
 
-build/nginx-proxy/concent-api-assets.tar: $(CONCENT_DIR)
+build/nginx-proxy/concent-api-assets.tar: $(CONCENT_DIR) build/virtualenv/
 	mkdir --parents "$(dir $@)"
 	package-builder/build-concent-api-assets.sh \
 		"$(CONCENT_DIR)"                        \
--- a/containers/package-builder/build-virtualenv.sh
+++ b/containers/package-builder/build-virtualenv.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -e
+
+requirements_path=$(realpath "$1")
+build_dir=$(realpath "$2")
+
+mkdir --parents "$build_dir/virtualenv/"
+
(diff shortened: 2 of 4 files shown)


containers/Makefile
```diff
@@ -68,6 +68,11 @@ postgresql-clean:
 concent-api-clean:
 	docker rmi $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) --force
 
+build/concent-api/RELEASE-VERSION: build/nginx-proxy/concent-api-assets.tar
+	build/virtualenv/bin/python3                       \
+	"build/virtualenv/src/golem-messages/version.py"
+	mv RELEASE-VERSION "build/concent-api"
```
cameel: The parameters should be indented with respect to the command.

But you could just keep it on a single line and it would be fine.

cameel: Use `$@` as the second argument of `mv`.

cameel: There should be a separate target for creating the virtualenv. Right now creating the virtualenv is only a side-effect in the step that produces `concent-api-assets.tar` and you can't count that it will always stay like this. If both need the virtualenv, it should be created explicitly.

containers/concent-api/Dockerfile
```diff
@@ -4,6 +4,7 @@ COPY install-dependencies.sh        /tmp/
 RUN  /tmp/install-dependencies.sh
 
 COPY requirements.lock              /srv/http/concent_api/
+COPY RELEASE-VERSION                /srv/http/
```
cameel: Don't add it explicitly. This file should be inside `concent-api.tar`.

containers/Makefile
```diff
@@ -68,6 +68,11 @@ postgresql-clean:
 concent-api-clean:
 	docker rmi $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) --force
 
+build/concent-api/RELEASE-VERSION: build/nginx-proxy/concent-api-assets.tar
```
cameel: It's `concent-api-assets.tar` that should depend on `RELEASE-VERSION`, not the other way around. This file should be included inside the archive.

containers/Makefile
```diff
@@ -68,6 +68,10 @@ postgresql-clean:
 concent-api-clean:
 	docker rmi $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) --force
 
+
+build/virtualenv/: $(CONCENT_DIR)
+	package-builder/build-virtualenv.sh "$(CONCENT_DIR)" "build"
```
cameel: This should depend only on `concent_api/requirements.lock`. Not the whole `CONCENT_DIR`.

cameel: Pass the path to requirements not `CONCENT_DIR` in a parameter.

containers/package-builder/build-virtualenv.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+concent_dir=$(realpath "$1")
+build_dir=$(realpath "$2")
+
+mkdir --parents "$build_dir/virtualenv/"
+cp --recursive "$concent_dir/concent_api/requirements.lock" "$build_dir/virtualenv"
```
cameel: You don't need `--recursive` here. You're not copying a directory.

Actually you don't need to copy the requirements at all.

containers/Makefile
```diff
@@ -76,12 +80,14 @@ build/concent-api/requirements.lock: build/concent-api/concent-api.tar
 		--directory build/concent-api/           \
 		concent_api/requirements.lock
 
-build/concent-api/concent-api.tar: $(CONCENT_DIR)
+build/concent-api/concent-api.tar: $(CONCENT_DIR)  build/virtualenv/%
```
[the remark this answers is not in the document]
bartoszbetka: Yes, It's work, but I delete it.

PR #72, opened 2018-03-02
Add json errors and accept header for nginx-proxy and nginx-storage

### LINKED ISSUE #42: JSON error messages from nginx
Configure our nginx instances to always return errors in JSON, not HTML.

We have that working for HTTP 403 but 404, 413, 5xx and probably others are not covered. Check which errors are in HTML and make them all JSON.

bartoszbetka:
Resolves #42

### DIFF
--- a/kubernetes/config-maps/nginx-proxy/default.conf.j2
+++ b/kubernetes/config-maps/nginx-proxy/default.conf.j2
@@ -1,5 +1,10 @@
 resolver kube-dns.kube-system.svc.cluster.local;
 
+map $http_accept $extension {
+    default html;
+    application/json json;
+}
+
 server {
     listen 80;
 
@@ -7,6 +12,7 @@ server {
     gzip_min_length 1000;
     gzip_proxied    expired no-cache no-store private auth;
     gzip_types      text/plain text/css text/javascript application/xml;
+    include         snippets/error-pages.conf;
 
     location / {
         root    /srv/http/nginx-proxy;
--- a/kubernetes/config-maps/nginx-storage/default.conf.j2
+++ b/kubernetes/config-maps/nginx-storage/default.conf.j2
@@ -1,5 +1,10 @@
 resolver kube-dns.kube-system.svc.cluster.local;
 
+map $http_accept $extension {
+    default html;
+    application/json json;
+}
+
 server {
     listen 80;
 
@@ -9,6 +14,9 @@ server {
     gzip_types               text/plain text/css text/javascript application/xml;
     client_max_body_size     {{ max_upload_size }};
 
+    # NOTE: OpenResty uses /usr/local/openresty/nginx/, not /etc/nginx/ as the default config location.
+    # We have to use absolute path to include files from /etc/nginx/.
+    include                  /etc/nginx/snippets/error-pages.conf;
 
     set $gatekeeper_api_backend http://gatekeeper.default.svc.cluster.local;
 
@@ -18,7 +26,6 @@ server {
         if ($request_uri ~ "/download/(.*)") {
             set $download_url $1;
         }
-        error_page 401 = @error_401;
         alias /srv/storage/;
     }
 
@@ -33,7 +40,6 @@ server {
         auth_request /upload-auth/;
         auth_request_set $checksum $upstream_http_concent_file_checksum;
         auth_request_set $size     $upstream_http_concent_file_size;
-        error_page 401 = @error_401;
 
         access_by_lua_block {
             -- Add headers from gatekeeper response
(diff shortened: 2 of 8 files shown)


kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -37,4 +53,49 @@ server {
...
+    }
+
+    location @error_408 {
+        default_type application/json;
+        return 408 '{"error_message": "The server timed out waiting for the request", "error_code": "request-timeout"}';
+    }
+
+    location @error_500 {
+        default_type application/json;
+        return 500 '{"error_message": "The server occur unexpected error", "error_code": "internal-server-error"}';
```
cameel: Please use the standard messages. This should be for example `Internal server error`.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -195,8 +209,48 @@ server {
...
+    }
+
     location @error_401 {
         default_type application/json;
         return 401 '{"error_message": "Access denied.", "error_code": "unauthorized"}';
     }
+
+    location @error_403 {
+        default_type application/json;
+        return 403 '{"error_message": "Not permission for this page", "error_code": "forbidden"}';
```
cameel: This part repeats in both nginx instances. Extract repeating parts into a snipped and just include them in both. We don't want repeated code that can easily go out of sync.

cameel: Are these all errors that nginx can return? What about the ones not listed here? Is there a generic mechanism that can let us replace all the errors so that we're sure no non-JSON errors ever get returned?

bartoszbetka: Well, I don't find alternative solution to replace all non-JSON errors. I think that all most common errors are covered. In `nginx-storage` config are more errors included, which is cover possible errors related to uploading files. Are we want cover all possible error ?

cameel: >  In nginx-storage config are more errors included

Which one specifically?

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -8,7 +13,7 @@ server {
     gzip_proxied             expired no-cache no-store private auth;
     gzip_types               text/plain text/css text/javascript application/xml;
     client_max_body_size     {{ max_upload_size }};
-
+    include                  snippets/json-errors.conf;
```
cameel: Call this `error-pages.conf`. These won't necessarily be just JSON errors. If the client requests HTML via the `Accept` header, we should return HTML.

kubernetes/config-maps/nginx/json-errors.conf
```diff
@@ -0,0 +1,55 @@
+error_page 400 = @error_400;
```
cameel: Common codes missing: 405, 406, 415

kubernetes/services/nginx-storage.yml.j2
```diff
@@ -30,11 +30,17 @@ spec:
               name:       concent-storage
             - mountPath:  /etc/nginx/conf.d/
               name:       nginx-storage-config
+            - mountPath:  /usr/local/openresty/nginx/conf/snippets/
+              name:       nginx-json-errors
+              readOnly:   true
```
cameel: Add a comment explaining why we're putting this config into `/usr/` and not into `/etc/nginx`.

kubernetes/delete-config-maps.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash -e
 
+kubectl delete configmap nginx-json-errors    || true
```
cameel: Add the file to existing config maps instead of creating a new one. It's an integral part of the current config, not something independent.

EDIT: I see it's in a different map because you can't put it in `/etc/nginx/`. You should explain that in a comment.

Maybe you could still put it in `/etc/` and just import using an absolute path?

bartoszbetka: I put it in the same place as nginx-proxy `/ etc / nginx / snippets`, and in the ` default.conf` file, I used the absolute path to import it.

bartoszbetka: I put it in the same place as nginx-proxy `/etc/nginx/snippets`, and in the ` default.conf` file, I used the absolute path to import it.

kubernetes/config-maps/nginx/error-pages.conf
```diff
@@ -0,0 +1,70 @@
+error_page 400 = @error_400;
```
cameel: You added `location` for 405, 406 and 415 below but you forgot to add `error_page` for them here.

kubernetes/delete-config-maps.sh
cameel: You should call the map `nginx-config-snippets`.

PR #83, opened 2018-03-16
Change condition, now cluster support HTTPS by default

### LINKED ISSUE #80: Make HTTPS always available
Now, after #55 we have the choice between enabling and not enabling HTTPS. After giving it some thought I came to a conclusion that there's no reason not to have HTTPS always enabled. We should just be disabling plain HTTP when we do not want it.

- Rename `nginx_proxy_ssl` and `nginx_storage_ssl` to `nginx_proxy_allow_http` and `nginx_storage_allow_http`.
- Change conditions so that the SSL stuff is always enabled. Instead the variables should change only one thing: when true, nginx should handle HTTP traffic normally; when false, HTTP should redirect to HTTPS.
- Update the info about SSL secrets in README. Now the certificates are always needed. They cannot be empty.

This should simplify the deployment and also resolve the issue that HTTP certificates are always needed, even if HTTPS is not enabled.

bartoszbetka:
Resolves #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,6 @@ Put them in the following locations:
 - `concent-secrets/$cluster/nginx-storage-ssl.crt`
 - `concent-secrets/$cluster/nginx-storage-ssl.key`
 
-NOTE: These files are needed even if you disable SSL for a given cluster (by setting `nginx_proxy_ssl = false` or `nginx_storage_ssl = false` in `vars.yml`).
-If SSL is disabled for a cluster, the corresponding files can contain anything.
-They can even be empty.
-They will not be deployed to the cluster.
-
 #### Generating self-signed certificates
 
 It's best if your certificates are signed by a Certificate Authority (CA) because then it's possible for the client to verify their authenticity without having to know the public key ahead of time.
--- a/kubernetes/config-maps/nginx-proxy/default.conf.j2
+++ b/kubernetes/config-maps/nginx-proxy/default.conf.j2
@@ -5,7 +5,8 @@ map $http_accept $extension {
     application/json json;
 }
 
-{% if nginx_proxy_ssl %}
+{% if not nginx_proxy_allow_http %}
+# HTTP -> HTTPS redirect
 server {
     listen 80      default_server;
     listen [::]:80 default_server;
@@ -13,14 +14,16 @@ server {
     return 302     https://$server_name$request_uri;
 
 }
+{% endif %}
 
 server {
+    {% if nginx_proxy_allow_http %}
+    # HTTP
+    listen 80       default_server;
+    {% endif %}
+    # HTTPS
     listen 443      ssl http2 default_server;
     listen [::]:443 ssl http2 default_server;
-{% else %}
-server {
-    listen 80       default_server;
-{% endif %}
 
     gzip            on;
     gzip_min_length 1000;
--- a/kubernetes/config-maps/nginx-storage/default.conf.j2
+++ b/kubernetes/config-maps/nginx-storage/default.conf.j2
@@ -5,7 +5,8 @@ map $http_accept $extension {
     application/json json;
 }
 
-{% if nginx_storage_ssl %}
+{% if not nginx_storage_allow_http %}
+# HTTP -> HTTPS redirect
 server {
     listen 80      default_server;
     listen [::]:80 default_server;
@@ -13,14 +14,16 @@ server {
     return 302     https://$server_name$request_uri;
 
 }
+{% endif %}
 
 server {
+    {% if nginx_storage_allow_http %}
+    # HTTP
+    listen 80       default_server;
+    {% endif %}
+    # HTTPS
     listen 443      ssl http2 default_server;
     listen [::]:443 ssl http2 default_server;
-{% else %}
-server {
-    listen 80                default_server;
-{% endif %}
 
     gzip                     on;
     gzip_min_length          1000;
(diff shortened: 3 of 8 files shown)


README.md
```diff
@@ -64,9 +64,8 @@ Put them in the following locations:
 - `concent-secrets/$cluster/nginx-storage-ssl.crt`
 - `concent-secrets/$cluster/nginx-storage-ssl.key`
 
-NOTE: These files are needed even if you disable SSL for a given cluster (by setting `nginx_proxy_ssl = false` or `nginx_storage_ssl = false` in `vars.yml`).
-If SSL is disabled for a cluster, the corresponding files can contain anything.
-They can even be empty.
+NOTE: These files are needed even if you allow connection without ssl for a given cluster (by setting `nginx_proxy_allow_http = true` or `nginx_storage_allow_http = true` in `vars.yml`).
+They can not be empty.
```
cameel: This paragraph is no longer necessary. Remove it.

It was there because it's counter-intuitive that the cert files need to be provided even when SSL is disabled. That's no longer true.

kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -5,8 +5,11 @@ map $http_accept $extension {
     application/json json;
 }
 
-{% if nginx_proxy_ssl %}
-server {
+{% if nginx_proxy_allow_http %}
+ server {
+    listen 80       default_server;
+{% else %}
```
cameel: This condition enables only HTTP when `nginx_proxy_allow_http` is true. What about HTTPS? HTTPS should be enabled in that case too.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -5,8 +5,11 @@ map $http_accept $extension {
     application/json json;
 }
 
-{% if nginx_storage_ssl %}
-server {
+{% if nginx_storage_allow_http %}
+ server {
+    listen 80       default_server;
+{% else %}
```
cameel: Same as above. This does not enable HTTPS.

kubernetes/create-config-maps.sh.j2
```diff
@@ -4,19 +4,19 @@
 kubectl create configmap nginx-config-snippets                          \
     --from-file=error-pages.conf=config-maps/nginx/error-pages.conf
 
-{% if nginx_storage_ssl %}
+{% if not nginx_storage_allow_http %}
 kubectl create secret generic nginx-storage-secrets                              \
```
cameel: You should remove all conditions from this file. SSL is now always enabled.

kubernetes/services/nginx-storage.yml.j2
```diff
@@ -34,13 +34,13 @@ spec:
             - mountPath:  /etc/nginx/snippets/
               name:       nginx-config-snippets
               readOnly:   true
-            {% if nginx_storage_ssl %}
+            {% if not nginx_storage_allow_http %}
             - mountPath:  /etc/ssl/secrets/
```
cameel: You should remove all conditions from this file. SSL is now always enabled.

README.md

kubernetes/services/nginx-storage.yml.j2
bartoszbetka: Done.

kubernetes/create-config-maps.sh.j2
bartoszbetka: Done.

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-proxy/default.conf.j2
bartoszbetka: Done.

kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -5,22 +5,22 @@ map $http_accept $extension {
     application/json json;
 }
 
-{% if nginx_proxy_ssl %}
-server {
+{% if not nginx_proxy_allow_http %}
+ server {
```
cameel: Why the extra space here?

cameel: Could you merge the `server` blocks and put all conditions inside it. I think it would make the config a little simpler.

bartoszbetka: Done.

bartoszbetka: I think we can do something like this: https://serverfault.com/a/474345, but I don't know, if this solution will simplify config.

cameel: OK, nevermind. It can stay as is. I thought that we can join the blocks without adding the extra `if`.

kubernetes/config-maps/nginx-proxy/default.conf.j2
```diff
@@ -5,22 +5,22 @@ map $http_accept $extension {
     application/json json;
 }
 
-{% if nginx_proxy_ssl %}
+{% if not nginx_proxy_allow_http %}
 server {
     listen 80      default_server;
```
cameel: We talked about it in person but I'll add it here to so that I don't forget. Just a minor change:

Please add a comments saying `HTTP -> HTTPS redirect`, `HTTP` and `HTTPS` before each part. this will make it a bit clearer.

bartoszbetka: Done.

PR #85, opened 2018-03-19
Add file with list of containers and packages versions

### LINKED ISSUE #78: Single list of container and package versions
We need a file where we could keep versions of base Docker images, images used in Kubernetes configs, linux packages and anything else that's hard-coded in the deployment repo. We want versions hard-coded to avoid situations where the version can change during deployment but we need to do it in a way that does not make it hard to find and update those versions regularly.

Things to do:
- Create `containers/versions.yml`. Define versions in it, like this:
    ``` yaml
    debian_version:   "9.4"
    postgres_version: "10.3-alpine"
    nginx_version:    "1.13.9-alpine"
    python_version:   "3.6"
    ```
- Include this file as one of the context files when rendering templates. Use these variables in `Dockerfile`s and scripts.
- Versions used in Ansible scripts for the build server should be there too. Python version, docker version, etc. You can add the version file as any other file with variables.
- Add versions in `Dockerfiles` and scripts that do not have versions hard-coded now.
    - You can also update everything else to the latest version (but in a separate commit).
- Put any versions used in Kubernetes configs in `versions.yml` too. Then include this file as context when building Kubernetes scripts.

Things to consider
- It might be also a good idea to extract whole package lists from scripts into `versions.yml`.

bartoszbetka:
Resolves #78


containers/Makefile
```diff
@@ -112,8 +112,14 @@ build/%: %
 	cp --archive --parents "$<" build/
 
 # Templates
-build/%: %.j2 config.yml ../tools/render-jinja-template.sh
-	../tools/render-jinja-template.sh config.yml "$<" "$@"
+build/%: %.j2 build/vars-merged-with-config.yml ../tools/render-jinja-template.sh
+	../tools/render-jinja-template.sh build/vars-merged-with-config.yml "$<" "$@"
+
+build/vars-merged-with-config.yml: versions.yml config.yml ../tools/merge-yaml.py
```
cameel: Please call it `config-with-versions.yml`. You're not using `vars.yml` here.

containers/Makefile
```diff
@@ -112,8 +112,14 @@ build/%: %
 	cp --archive --parents "$<" build/
 
 # Templates
-build/%: %.j2 config.yml ../tools/render-jinja-template.sh
-	../tools/render-jinja-template.sh config.yml "$<" "$@"
+build/%: %.j2 build/vars-merged-with-config.yml ../tools/render-jinja-template.sh
+	../tools/render-jinja-template.sh build/vars-merged-with-config.yml "$<" "$@"
```
cameel: Please break this command into multiple lines because it's starting to get hard to read.

containers/versions.yml
```diff
@@ -0,0 +1,7 @@
+debian_version:     "9.4"
+geth_version:       "v1.8.2"
+gunicorn_version:   "19.7.1"
+nginx_version:      "1.13.9-alpine"
+openresty_version:  "1.13.6.1-1-alpine-fat"
+postgres_version:   "10.3-alpine"
```
cameel: I think that maybe it would be better to not include the `-alpine` part in the version and hard-code it in `Dockerfile`s instead. The `Dockerfiles` are built with the assumption that the underlying system is alpine so if someone changes it here they won't work.

concent-builder/configure.yml
```diff
@@ -76,8 +77,8 @@
         - name:    Install dependencies from testing
```
cameel: Please hard-code versions of packages that come from PPAs (and not official repositories): docker, kubectl, Google Cloud SDK

containers/versions.yml
bartoszbetka: That's good idea, but I think we should also change variables e.g from `postgres_version` to `postgres_alpine_version`. When we will want some changes to version, we will not have to wonder what linux version of the image we are using.

containers/Makefile

containers/Makefile

containers/versions.yml
bartoszbetka: Done.

concent-builder/configure.yml
bartoszbetka: Done.

containers/concent-api/install-concent-api.sh.j2
```diff
@@ -1,11 +1,10 @@
...
-python_version=3.6
 
 # Create virtualenv and Django app's dependencies
-virtualenv --python python$python_version "$virtualenv_dir"
+virtualenv --python python{{ python_version }} "$virtualenv_dir"
 source "$virtualenv_dir/bin/activate"
 pip install --upgrade pip
 pip install --requirement "/srv/http/concent_api/requirements.lock"
-pip install gunicorn
+pip install gunicorn=={{ gunicorn_version }}
```
cameel: Please put arguments that take values from `{{ }}` expressions in quotes. They could contain spaces.

containers/package-builder/build-concent-api-docs.sh.j2
```diff
@@ -10,7 +10,7 @@ cp --recursive "$concent_dir/docs" "$build_dir/concent-api-docs"
 
 cd "$build_dir/concent-api-docs/docs/"
 
-virtualenv --python python3 "$build_dir/docs-virtualenv/"
+virtualenv --python python{{ python_version }} "$build_dir/docs-virtualenv/"
```
cameel: Please put `python{{ python_version }}` in quotes.

Do this here an in other places in this pull request.

containers/concent-api/install-concent-api.sh.j2
bartoszbetka: Done.

containers/package-builder/build-concent-api-docs.sh.j2
bartoszbetka: Done.

PR #92, opened 2018-03-27
Add deployment and service with rabbitmq on the cluster

### LINKED ISSUE #88: Deployment and service for RabbitMQ on the storage cluster
- Create a deployment that uses the official [rabbitmq docker image](https://hub.docker.com/_/rabbitmq/)
    - Configure it to run one instance
- Provision a small persistent disk for the pod so that it can store its database in it (check how much storage it usually needs). Mount it as a volume in the right directory.
- Create a service that makes the pod available to other services on the storage cluster.

bartoszbetka:
Resolves #88


kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,70 @@
...
+  name: rabbitmq
+spec:
+  replicas:             1
+  revisionHistoryLimit: 1
+  template:
+    metadata:
+      labels:
+        run: rabbitmq
+    spec:
+      hostname: rabbitmq-1
```
cameel: Change it to just `rabbitmq`. We're not going to have more than one node per cluster.

Also, add a comment explaining why we need a specific hostname and how RabbitMQ uses hostname.

kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,70 @@
...
+            - mountPath:  /srv/storage/mnesia/
+              name:       queue-storage
+          livenessProbe:
+            exec:
+              command:
+              - rabbitmqctl
+              - status
+            initialDelaySeconds: 120
+            timeoutSeconds: 5
+            failureThreshold: 6
```
cameel: How were these values established?

Also, please align this.

kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,70 @@
...
+        - name: queue-storage
+          gcePersistentDisk:
+            pdName: {{ queue_disk }}
+            fsType: ext4
+      # securityContext allows you to add permission for pod and mounted volumes to specific group.
+      # We need this permission to create files and directories on mounted persistent disk.
+      # Rabbitmq user has reserved 999 ID
+      # We use this user when creating files and directories.
+      securityContext:
+        fsGroup: 999
```
cameel: 1) > Rabbitmq user has reserved 999 ID

    `fsGroup` is the ID of the group, not of the user

2) Are you sure the ID is 999?

    Here's how they create the user and the group: https://github.com/docker-library/rabbitmq/blob/master/3.7-rc/alpine/Dockerfile#L4
    ``` Dockerfile
    # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
    RUN addgroup -S rabbitmq && adduser -S -h /var/lib/rabbitmq -G rabbitmq rabbitmq
    ```

    They do not have it hard-coded to 999 and their instructions will not result in that ID. You can verify with:

    ``` bash
    docker run \
        --interactive \
        --tty \
        --rm \
        alpine:3.7 \
            sh -c \
                "addgroup -S rabbitmq && \
                adduser -S -h /var/lib/rabbitmq -G rabbitmq rabbitmq && \
                cat /etc/passwd | grep rabbitmq && \
                cat /etc/group | grep rabbitmq"
    ``` 

    The output is:
    ```
    rabbitmq:x:100:101:Linux User,,,:/var/lib/rabbitmq:/bin/false
    rabbitmq:x:101:rabbitmq
    ```    
    So their user has ID `100` and group has ID `101`.

kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,70 @@
...
+  replicas:             1
+  revisionHistoryLimit: 1
+  template:
+    metadata:
+      labels:
+        run: rabbitmq
+    spec:
+      hostname: rabbitmq-1
+      containers:
+        - image:           rabbitmq:3.7.4
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,70 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: rabbitmq
+spec:
+  replicas:             1
+  revisionHistoryLimit: 1
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,70 @@
...
+              command:
+              - rabbitmqctl
+              - status
+            initialDelaySeconds: 10
+            timeoutSeconds: 3
+            periodSeconds: 5
+      volumes:
+        - name: queue-storage
+          gcePersistentDisk:
+            pdName: {{ queue_disk }}
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/services/rabbitmq.yml.j2
bartoszbetka: Done. Values are from `kubernetes` github repository: https://github.com/kubernetes/charts/blob/master/stable/rabbitmq/values.yaml

kubernetes/services/rabbitmq.yml.j2
bartoszbetka: The output from rabbitmq pod:
`/etc/passwd`:
```
rabbitmq:x:999:999::/var/lib/rabbitmq:
```
`/etc/group`:
```
rabbitmq:x:999:

```

cameel: But how? Did I look at the wrong Docker image?

bartoszbetka: Yes, you are looking at a wrong image, we use this: https://github.com/docker-library/rabbitmq/blob/master/3.7/debian/Dockerfile

cameel: OK, so:
1) Why not use the alpine image?
2) Definitely add a comment that says that this ID depends on the base image. That's not obvious.

bartoszbetka: 1. I don't remember exactly, but I think the problem was related with feature and stability:
- https://docs.docker.com/samples/library/rabbitmq/#enabling-hipe
- https://docs.docker.com/samples/library/rabbitmq/#rabbitmqalpine
2. Ok, I added a comment.

kubernetes/services/rabbitmq.yml.j2
cameel: OK. It's fine as a starting point then.

In general, please try to use more comments explaining your reasoning in the configs.

kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,76 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: rabbitmq
+spec:
+  replicas:             1
```
cameel: Please add a comment explaining why it's set to 1:

RabbitMQ is stateful so we can't simply spin up several replicas without ending up with separate broker instances with separate queues. For now this has to be hard-coded to 1.

kubernetes/services/rabbitmq.yml.j2
```diff
@@ -0,0 +1,76 @@
...
+  template:
+    metadata:
+      labels:
+        run: rabbitmq
+    spec:
+# The default name of the rabbitmq node is the hostname whose name in kubernetes is the name of the nodes.
+# This name is random and change after restarting, deleting, etc.
+# This behavior causes creates a new database for the new node, so we can not recover messages and queues.
+# We need a constant name of rabbitmq node, so we can read messages and queues from an existing database.
+# Adding hostname field in the kubernetes deployment, change host name of the pod.
```
cameel: It's fine but please reword it:

The name of the RabbitMQ node (RABBITMQ_NODENAME) is by default the same as the hostname, which in Kubernetes is the same in every container running on a single Kubernetes node (and is the same as the hostname of that node). This name is random and changes when the pod is restarted or deleted and recreated. This behavior results in a new, empty Mnesia database being created when we recreate the pod and messages from the previous database not being recovered. For that reason we need to explicitly set the hostname for the pod and keep it unchanged between deployments.

kubernetes/services/rabbitmq.yml.j2
cameel: In that case let's switch to Alpine. The warning is only relevant if we would want to add some software that heavily depends on GNU libc to the image. We're not adding anything so for us alpine is fine (at least for now).

kubernetes/services/rabbitmq.yml.j2
bartoszbetka: Done.

kubernetes/services/rabbitmq.yml.j2
bartoszbetka: Done.

PR #102, opened 2018-04-04
Add broker_url variable which allow connection from celery to rabbitmq

### LINKED ISSUE #100: Configure Celery
In order to deploy the code with https://github.com/golemfactory/concent/issues/197 we need to configure Celery in `local_settings.py`.

- Configure it for Concent API. Gatekeeper does not need it. Verifier and Conductor should use the same Celery configuration in the future.
- Connect to the RabbitMQ server deployed in #88.
- Right now the only thing you need to configure should be the broker URL. Let me know in a comment if you think we need anything else.

bartoszbetka:
Resolves #100


bartoszbetka:
Yes, it works too, but when the `rabbitmq` is restarted or some error occurs, the `celery` will not be able to connect with it, before rebooting itself.

bartoszbetka:
There is some misunderstanding here. I assumed that the celery cache IP just like nginx, but it doesn't do this, so the `rabbitmq` name works the same as` rabbitmq.default.svc.cluster.local`

PR #104, opened 2018-04-06
Refactor concent settings - separation repeating parts of code

bartoszbetka:


kubernetes/config-maps/concent/admin_settings.py.j2
```diff
@@ -0,0 +1,5 @@
+ADMINS = [
+{% for admin in admins %}
+    ("{{ admin.name }}", "{{ admin.mail }}"),
+{% endfor %}
```
cameel: This is really a part of the e-mail settings. Please merge those two files.

kubernetes/config-maps/concent/email_settings.py.j2
```diff
@@ -0,0 +1,9 @@
+from ..secrets.secrets import EMAIL_PASSWORD
+
+EMAIL_HOST = "{{ email.host }}"
+EMAIL_PORT = {{ email.port }}
```
cameel: Please move `RAVEN_CONFIG` to this file too. Anything related to error reporting/Sentry should be here.

kubernetes/create-config-maps.sh.j2
```diff
@@ -35,4 +38,6 @@ kubectl create configmap concent-api-settings                                \
 
 kubectl create configmap gatekeeper-settings                               \
     --from-file=local_settings.py=config-maps/gatekeeper/local_settings.py \
+    --from-file=admin_settings.py=config-maps/concent/admin_settings.py    \
+    --from-file=email_settings.py=config-maps/concent/email_settings.py    \
```
cameel: I think we should add `timing_settings` to gatekeeper too for consistency.

Do it in a separate commit of course - because it's not just refactoring but a change in behavior.

kubernetes/config-maps/concent/email_settings.py.j2
bartoszbetka: There is a problem with moving `RAVEN CONFIG`, because we added tag with hard coded name of container.

kubernetes/config-maps/concent/admin_settings.py.j2
bartoszbetka: Done.

kubernetes/create-config-maps.sh.j2
bartoszbetka: Done.

kubernetes/config-maps/concent/email_settings.py.j2
cameel: In that case define an extra settings module called `constants.py` with a setting called `KUBERNETES_COMPONENT_NAME`. Then import and use it in all places where we have this hard-coded in settings right now.

PR #115, opened 2018-04-23
Delete storage database from concent-api settings

bartoszbetka:


kubernetes/config-maps/concent-api/local_settings.py.j2
```diff
@@ -19,6 +19,8 @@ DATABASES['default']['PASSWORD'] = DATABASE_PASSWORD
 DATABASES['default']['HOST']     = '{{ database_host }}'
 DATABASES['default']['PORT']     = '5432'
 
+del DATABASES["storage"]
```
cameel: Looks OK, but you should add a comment explaining why we're doing that:

Concent has two databases in its default config but only one is used at a time and they may not be even on the same cluster. In order to work with only one we have to remove the other from the config.

bartoszbetka: Done

PR #122, opened 2018-04-27
 Add script which wait for pod running, before create another one 

bartoszbetka:


kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
+#! /bin/bash -e
+
+pod_label=$1
+statefulset=$2
```
cameel: Please add a short description that says what the parameters are and how to run the script.

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Statefulset starts the pods individually, but with the convention `{statefulset-name}-{0-n}`
+# The second member {0-n} corresponds to the amount of pods
+# We have to manually defined name of the pods based of amount of replicas.
+if [[ "$statefulset" == "true" ]]; then
+    pods_name=()
+    for i in `seq 0 $(($(kubectl get statefulset $pod_label -o jsonpath='{.spec.replicas}')-1))`; do
```
cameel: Please use a helper variable for readability:
``` bash
replica_count="$(kubectl get statefulset $pod_label --output jsonpath='{.spec.replicas}')"
```

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
+#! /bin/bash -e
+
+pod_label=$1
+statefulset=$2
+
+# Statefulset starts the pods individually, but with the convention `{statefulset-name}-{0-n}`
+# The second member {0-n} corresponds to the amount of pods
+# We have to manually defined name of the pods based of amount of replicas.
+if [[ "$statefulset" == "true" ]]; then
+    pods_name=()
```
cameel: This holds multiple values so a name in plural would be more self-explanatory: `pod_names`

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
...
+    for i in `seq 0 $(($(kubectl get statefulset $pod_label -o jsonpath='{.spec.replicas}')-1))`; do
+        pods_name+=(pods/$pod_label-$i)
+    done
+    [ -z "$pods_name" ] && echo "The pod does not exist!" && exit 1
+else
+    pods_name=$(kubectl get pod -l "run=$1" -o name)
```
cameel: Please use long option names: `-l` -> `--selector`, `-o` -> `--output`.

cameel: This is a scalar so `pod_name` would be more more self-explanatory.

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
...
+
+    [ -z "$pods_name" ] && echo "The pod does not exist!" && exit 1
+
+    # Divide string to array by newline characters
+    readarray -t pods_name <<< "$pods_name"
```
cameel: Please reword:

``` bash
Create an array containing lines from $pods_name
```

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+for i in "${pods_name[@]}"; do
+    timeout              \
+        --verbose        \
+        70               \
```
cameel: Please put `70` on the same line as `timeout` for readability.

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
...
+        bash -c          \
+        "until kubectl get $i -o jsonpath='{.status.conditions[?(@.type==\"Ready\")].status}' | grep True ; do
```
cameel: Wouldn't a comparison be better than grepping it?

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
+#! /bin/bash -e
```
cameel: Please rename the script to `wait-until-ready.sh`.

kubernetes/check-pods.sh
cameel: Please make the timeout a parameter.

kubernetes/check-pods.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+for i in "${pods_name[@]}"; do
+    timeout              \
+        --verbose        \
+        70               \
+        bash -c          \
+        "until kubectl get $i -o jsonpath='{.status.conditions[?(@.type==\"Ready\")].status}' | grep True ; do
+            sleep 1 ;
+         done"
```
cameel: Please extract this part into a separate script file. This is a bit dense which will be a problem if we want to extend it later.

kubernetes/check-pods.sh
cameel: You're relying on the `run` label here. You should explain at the beginning of the script that it only works with pods that have this label.

kubernetes/check-pods.sh
cameel: Instead of requiring this parameter maybe it would be better to check if there's either a pod or stateful set with that name?

If you want to keep the parameter please use a value that's more self-explanatory. When you call the script as `./check-pods.sh verifier true` it's hard to tell what this `true` means. Maybe add an option like `--stateful-set`? Or `--type=statefulset|pod`?

kubernetes/create-services.sh
```diff
@@ -2,10 +2,19 @@
 
 ./create-config-maps.sh
 
+kubectl create --record --filename services/verifier.yml
+./check-pods.sh "verifier" "true"
 kubectl create --record --filename services/geth.yml
+./check-pods.sh "geth"
 kubectl create --record --filename services/rabbitmq.yml
+./check-pods.sh "rabbitmq"
```
cameel: We don't need to wait after each `create` separately. We're unnecessarily delaying the deployment. Wait only when you need to. You can move all the other waits after the `create`s.

Might be convenient to make it possible to pass multiple pod names to the `./check-pods.sh` script.

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: Done.

bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: It's not possible to do that. From `timeout --help` command:
> Usage: timeout [OPTION] DURATION COMMAND [ARG]...

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: Could you be more specific ? You want an input parameter that contains the time or the regular parameter.

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
cameel: Fine. Leave it as is.

cameel: I want it to be a parameter of the `check-pods.sh` script so that you can do:

``` bash
./check-pods.sh rabbitmq false 120
```

or, even better (you could use something like [getopt](http://aplawrence.com/Unix/getopts.html) for parsing arguments):
``` bash
./check-pods.sh rabbitmq --stateful-set=false --timeout=120
```

kubernetes/create-services.sh
bartoszbetka: Are you sure ? This script was created to wait for the creation of a statefulset pods before creating the deployment pods. This behavior allows to the best distribution of resources. When we will implement your version, all deployment pods will be create before the statefulset pods, so there will be no resources for statefulset pods

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/check-pods.sh
bartoszbetka: Done.

kubernetes/create-services.sh
bartoszbetka: Done.

cameel: That's why I said "wait only when you need to". Create stateful sets, wait for them, then create everything else, wait for it. No need to wait after every pod separately when others could be created at the same time.

kubernetes/check-for-pod-ready.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash -e
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/wait-until-ready.sh
```diff
@@ -0,0 +1,34 @@
+#! /bin/bash -e
+# This script works only with the pods containing `run` label.
+
+# The variable contain the pod label name
+pod_label=$1
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/wait-until-ready.sh
```diff
@@ -0,0 +1,34 @@
...
+# The variable contains the time after which the timeout will be returned
+timeout=$2
+
+# Statefulset starts the pods individually, but with the convention `{statefulset-name}-{0-n}`
+# The second member {0-n} corresponds to the amount of pods
+# We have to manually defined name of the pods based of amount of replicas.
+check_kind_of_pod="$(kubectl get deployment --selector "run=$pod_label" --output name)"
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/wait-until-ready.sh
```diff
@@ -0,0 +1,34 @@
...
+else
+    pod_name=$(kubectl get pod --selector "run=$1" --output name)
+
+    [ -z "$pod_name" ] && echo "The pod does not exist!" && exit 1
+
+    # Create an array containing lines from $pod_name.
+    readarray -t pod_names <<< "$pod_name"
+fi
+
+for i in "${pod_names[@]}"; do
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/create-services.sh
```diff
@@ -2,10 +2,19 @@
...
 kubectl create --record --filename services/rabbitmq.yml
 kubectl create --record --filename services/concent-api.yml
 kubectl create --record --filename services/gatekeeper.yml
 kubectl create --record --filename services/conductor.yml
 kubectl create --record --filename services/nginx-storage.yml
 kubectl create --record --filename services/nginx-proxy.yml
+./wait-until-ready.sh "geth" "60"
+./wait-until-ready.sh "rabbitmq" "60"
+./wait-until-ready.sh "concent-api" "30"
+./wait-until-ready.sh "gatekeeper"  "30"
```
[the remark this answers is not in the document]
bartoszbetka: Done.

concent-builder/build-test-and-push.yml
```diff
@@ -64,12 +64,12 @@
...
-              --rm                                                                                                                                            \
-              --volume {{ build_dir }}/concent-deployment/kubernetes/build/run-unit-tests.sh:/usr/local/bin/run-unit-tests.sh:ro                              \
-              --volume {{ build_dir }}/concent-deployment/kubernetes/build/docker_testing.py:/srv/http/concent_api/concent_api/settings/docker_testing.py:ro  \
-              --network {{ cluster }}-{{ image_prefix }}unittest-network concent-concent-api:latest /usr/local/bin/run-unit-tests.sh
+   #     - name:    Run concent-api container and django unit tests
+   #       command: docker run                                                                                                                                 \
+   #           --rm                                                                                                                                            \
+   #           --volume {{ build_dir }}/concent-deployment/kubernetes/build/run-unit-tests.sh:/usr/local/bin/run-unit-tests.sh:ro                              \
```
cameel: Why is this commented out?

bartoszbetka: It was accidentally committed. I know this, I will remove it before merging.

PR #123, opened 2018-05-07
Add upload information for conductor to nginx-storage

### LINKED ISSUE #89: Sending upload notifications from nginx-storage to Conductor
When an upload operation finishes successfully and the file is already in the target location in the storage, we want nginx-storage to make a request to a predefined service on the storage cluster. That service will then be responsible for triggering other actions for that file, for example adding a task to the work queue.

**Example:** When a client uploads a file called `blender/result/ef0dc1/ef0dc1.zzz523.zip` to `https://storage.concent.golem.network/upload/`, nginx storage should make a request equivalent to the following curl command:
``` bash
curl -X POST http://conductor.default.svc.cluster.local/conductor/report-upload/blender/result/ef0dc1/ef0dc1.zzz523.zip
```

The `http://conductor.default.svc.cluster.local/conductor/report-upload/` part should be configurable via a variable called `upload_report_url` in `vars.yml`.

If the `report-upload` request fails, nginx should return HTTP 502 to the client but leave the file intact (we don't want to mess with other upload requests for the same file which theoretically could be running in parallel).

If the upload request fails in any way, nginx-storage should not make the `report-upload` request. It should be performed if and only if the uploaded file actually ends up in our storage.

bartoszbetka:
Resolves #89


containers/nginx-storage/install-dependencies.sh.j2
```diff
@@ -21,11 +21,22 @@ ln -s /usr/lib/libcrypto.so.41 /usr/lib/libcrypto.so
...
-    --branch v1.4.0                                  \
+    --branch {{ nginx_big_upload_version }}          \
     --depth  1
 
 rm -rf /opt/nginx-big-upload/.git
 
+# Github repository with lua-resty-http module
+git clone                                            \
+    https://github.com/pintsized/lua-resty-http.git  \
+    /opt/lua-resty-http                              \
```
cameel: Adding a new dependency should be always done in a separate commit.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
 
             os.rename("/srv/storage/" .. tmp_file_name, "/srv/storage/" .. header_path_with_filename)
             os.remove("/srv/storage/" .. tmp_file_name .. ".shactx")
+
+            -- Send confirmation to the conductor after uploaded file.
```
cameel: `after uploaded file` -> `when a file is uploaded`.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
...
+            local http = require "resty.http"
+            local httpc = http.new()
+            local res, err = httpc:request_uri("{{ upload_report_url }}" .. header_path_with_filename, {
```
cameel: `res` is `result` or `response`? Please use full names.

`err` -> `conductor_error` or `upload_report_error`
`res` -> `conductor_response` or `upload_report_response`

cameel: Please add a comment somewhere near the definition of `upload_report_url` that it must contain the path. E.g. `http://conductor.default.svc.cluster.local/`, not `http://conductor.default.svc.cluster.local`.

Otherwise someone could give us a filename like `.my.own.domain.com/my-malicious-endpoint` so that we end up making a request to `http://conductor.default.svc.cluster.local.my.own.domain.com/my-malicious-endpoint/` which is outside the cluster!

It would be best if we could have some sanity check for `upload_report_url` so that's impossible to make such a mistake. But at least add a comment so that someone setting it is aware of it.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
...
+            method = "POST",
+            })
```
cameel: When splitting code into multiple lines do it like this:

``` lua
local res, err = httpc:request_uri(
    "{{ upload_report_url }}" .. header_path_with_filename,
    {method = "POST"},
)
```

So, each parameter on a separate line and one level of indentation.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
...
+
+            if (res.status < 200 or res.status >= 300) then
+                ngx.status = 502
+                error_response = '{"error_message":"Failed to report upload. Conductor responded with status {' .. res.status .. '}", "conductor_message":"' .. res.body .. '"}'
```
cameel: There should be a space after `:`.

cameel: Please include `error_code` in the response.

You could also split it into multiple lines for readability.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
...
+                ngx.say(error_response)
+                -- Return response body in nginx log for debugging purposes
+                ngx.log(ngx.ERR, error_response)
+                return ngx.exit(502)
```
cameel: Would it be possible to do `return ngx.exit(ngx.status)` here? You would not have to repeat the HTTP status twice.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
...
+
             -- Nginx big upload module adds headers with checksum and temporary file name to its response.
             -- We don't want to return them to the client.
             -- FIXME: Setting it to an empty string does not remove the headers. For now we set it to a
             -- single space to work around this but we should really just find a way to remove those headers.
             ngx.header["X-SHA1"]       = " "
             ngx.header["X-Session-Id"] = " "
             ngx.header.content_type    = "application/json; charset=utf-8"
-            ngx.say('{"message": "File was successfully uploaded"}')
+            successfully_response      = '{"message": "File was successfully uploaded"}'
```
cameel: `successfully_response` -> `successful_response`

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
...
+            ngx.say(successfully_response)
+            ngx.log(ngx.NOTICE, successfully_response)
```
cameel: Adding more logging also could be a separate commit.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: What happens if the connection fails? For example if network is down or something is blocking the connection?

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,14 +208,34 @@ server {
...
             os.remove("/srv/storage/" .. tmp_file_name .. ".shactx")
+
+            -- Send confirmation to the conductor after uploaded file.
+            local http = require "resty.http"
+            local httpc = http.new()
+            local res, err = httpc:request_uri("{{ upload_report_url }}" .. header_path_with_filename, {
+            method = "POST",
+            })
+
+            if (res.status < 200 or res.status >= 300) then
```
cameel: This might be more readable as `not (200 <= response.status and response.status < 300)`

containers/nginx-storage/install-dependencies.sh.j2
```diff
@@ -21,11 +21,22 @@ ln -s /usr/lib/libcrypto.so.41 /usr/lib/libcrypto.so
 git clone                                            \
     https://github.com/pgaertig/nginx-big-upload.git \
     /opt/nginx-big-upload                            \
-    --branch v1.4.0                                  \
+    --branch {{ nginx_big_upload_version }}          \
```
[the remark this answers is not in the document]
bartoszbetka: Done.

containers/nginx-storage/install-dependencies.sh.j2
bartoszbetka: Done.

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: According to https://github.com/pintsized/lua-resty-http#request, `res` is the result of successful request, so also include error responses. Do we want to change the names to those you proposed ?

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: So, that's not exactly a point. Second argument (dict) contains all parameters. Below correctly version.
```
local res, err = httpc:request_uri(
    "{{ upload_report_url }}" .. header_path_with_filename,
    {
        method = "POST",
    }
)
```
I think that style from documentation is more understandable:
```
local res, err = httpc:request_uri("{{ upload_report_url }}" .. header_path_with_filename, {
    method = "POST",
})
```

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: Well, I don't remember if I tested your solution, but everywhere in this file we use  convention `return ngx.exit(number)`

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: Done

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: There are now three commits for 36 line of code. Are you sure that we want add next commit with one or two line of code?

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: I forgot about it. I will add a condition.
```
            if not res then
                ngx.status = 500
                ngx.say('{"error_message": "failed to request:' .. err .. '", "error_code": "failed request"}')
                return ngx.exit(500)
            end

```

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: Yes, if they're correct.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: If there's only one element in a dict, it's fine to put it on a single line.

Those other two versions you gave are also OK. I'd still probably choose mine but If you want to use one of them that's fine too.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: Yes, I know. And it had never sit right with me because we're (seemingly unnecessarily) repeating the code twice. I did not see a better solution back then but now I have realized that we could just do what I proposed above. If that works then I think we should start doing it from now on.

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: I added a comment in `concent-deployment-values` repositories in `var.yml` file.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: Yes. It's not about the size. Even a single-line change is a good commit if it's something unrelated to the other lines.

You make a separate commit because you want to show that it's a single, coherent change, not because your change is too big.

You should generally make smaller commits.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: Is `res` a boolean value? If not, then this condition should be more specific.

Also it should be more like `"HTTP request to conductor failed: "`

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2

bartoszbetka: According to https://github.com/pintsized/lua-resty-http#request, return table or nil. Condition above is from documentation example: https://github.com/pintsized/lua-resty-http#synopsis

bartoszbetka: Changed `error_message`

kubernetes/config-maps/nginx-storage/default.conf.j2

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: It's the same story as with `if x` vs `if x is not None` in Python. Some people like it because it's shorter to type but our convention is to check conditions explicitly - there are corner cases (at least in Python) that can easily trip you up if you're not careful.

So this should be something like `if res != nil`.

kubernetes/config-maps/nginx-storage/default.conf.j2
```diff
@@ -207,6 +207,29 @@ server {
...
+                ngx.status = 500
+                ngx.say('{"error_message": "HTTP request to conductor failed:' .. conductor_error .. '", "error_code": "failed-conductor-request"}')
+                return ngx.exit(ngx.status)
+            end
+
+            if not (200 <= conductor_response.status and conductor_response.status < 300) then
+                ngx.status = 502
+                error_response = '{"error_message": "Failed to report upload. Conductor responded with status {' .. conductor_response.status .. '}", ' ..
+                '"error_code": "failed-upload-report", ' ..
+                '"conductor_message": "' .. conductor_response.body .. '"}'
```
cameel: I got confused for a moment when I looked at this. I now see why. These are not separate statements but parts of a single string concatenation. There should be some extra indentation or parentheses to make this clear. Maybe like this?

``` lua
error_response = '{' ..
    '"error_message": "Failed to report upload. Conductor responded with status {' .. conductor_response.status .. '}", ' ..
    '"error_code": "failed-upload-report", ' ..
    '"conductor_message": "' .. conductor_response.body .. '"' ..
'}'
```

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel:
``` patch
-# Address of conductor endpoint for sending notification after successfully upload used by nginx-storage
+# Address of conductor endpoint for sending notification when a file is uploaded used by nginx-storage
+# The string with url path must end with slash, for example "http://conductor.default.svc.cluster.local/conductor/report-upload/",
+# not "http://conductor.default.svc.cluster.local/conductor/report-upload"
 upload_report_url: "http://conductor.default.svc.cluster.local/conductor/report-upload/"
```

That's not entirely true. It does not have to end with a slash. It's enough if it ends with a path (which starts with a slash).

bartoszbetka: How should it work without a slash? This variable is combined with a path to a file that looks like this `blender/result/ef0dc1/ef0dc1.zzz523.zip`, so when combinate it together you receive: `http://conductor.default.svc.cluster.local/conductor/report-uploadblender/result/ef0dc1/ef0dc1.zzz523.zip`

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: Ok.  I will change it, but I'm not sure if there is the same analogy in lua. This condition `if not res then` comes from the documentation, so it must work.

kubernetes/config-maps/nginx-storage/default.conf.j2

bartoszbetka:
The commits are in the sequence exactly as you say ?

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: I just realized that the condition should look like this: `if res == zero` or old version `if not res`. Which one we want to used ?

cameel:
> The commits are in the sequence exactly as you say ?

Oh, OK, you're right. I looked at https://github.com/golemfactory/concent-deployment/pull/123/commits and assumed that the order is the same as in https://github.com/golemfactory/concent-deployment/commits/master. But I see that it's reversed.

kubernetes/config-maps/nginx-storage/default.conf.j2
cameel: OK. You're right. It must end with a slash. I did not notice it before. If there's a path but no slash the attacker won't be able to change the domain but could still make nginx-storage connect to a different path. The risk that he'll be able to do anything malicious that way is much lower than when changing a domain name but still it's a valid reason to require a slash.

But please explain both cases in a comment.

kubernetes/config-maps/nginx-storage/default.conf.j2
bartoszbetka: This `if not res` condition is correct when `res` is nil, so we can choose between two options that I mentioned above.

PR #140, opened 2018-05-28
Add feature that pushes a unique version of containers into a cluster and requires a specific version of concent and golem

### LINKED ISSUE #134: Unique container versions
Currently our containers are tagged using the output of `git describe --always` which gives output like `v0.5.3-rc3-2-g3d1eb9a`. Those tags are not pushed to the Docker registry though and also they do not uniquely identify a particular build since they depend only on concent-deployment and specific Concent and Golem versions are not hard-coded in that repository.

- Make [kubernetes/Makefile#L57-L60](https://github.com/golemfactory/concent-deployment/blob/master/kubernetes/Makefile#L57-L60) construct the tag the same way [containers/Makefile#L49-L52](https://github.com/golemfactory/concent-deployment/blob/master/containers/Makefile#L49-L52) does.
    - This will embed the actual version in all scripts and Kubernetes configs.
- Add golem and concent versions to `containers/versions.yml` so that every version of concent-deployment is always associated with specific versions of those repositories.
    - The version should always be either a git tag or a git commit.
- Move cloning golem and concent repositories from the build playbook to the `Makefile`. Put them somewhere inside the `build/` directory and adjust all paths in the `Makefile` accordingly.
    - They should still be cloned from the local clone that we have in `/var/`.
    - You can also remove fetching concent-deployment-values completely (unless we're using it when building `kubernetes/` but I think we're always using the version uploaded from the local machine).

bartoszbetka:
Resolves #134

### DIFF
--- a/concent-builder/repositories.yml
+++ b/concent-builder/repositories.yml
@@ -3,7 +3,5 @@ repositories:
     url: https://github.com/golemfactory/concent.git
   concent-deployment:
     url: https://github.com/golemfactory/concent-deployment.git
-  concent-deployment-values:
-    url: https://github.com/golemfactory/concent-deployment-values.git
   golem:
     url: https://github.com/golemfactory/golem.git
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -37,16 +37,6 @@ VERIFIER_SOURCE :=                           \
     $(VERIFIER_SOURCE_FROM_CONCENT_API)
 
 
-ifndef CONCENT_DIR
-    CONCENT_DIR:= $(abspath ../../concent)
-endif
-$(info CONCENT_DIR=$(CONCENT_DIR))
-
-ifndef GOLEM_DIR
-    GOLEM_DIR:= $(abspath ../../golem)
-endif
-$(info GOLEM_DIR=$(GOLEM_DIR))
-
 ifndef IMAGE_VERSION
     IMAGE_VERSION := $(shell git describe --always)
 endif
@@ -72,11 +62,11 @@ postgresql: $(POSTGRESQL_SOURCE)
 	docker build --tag $(IMAGE_PREFIX)postgresql:$(IMAGE_VERSION) build/postgresql/
 	docker tag $(IMAGE_PREFIX)postgresql:$(IMAGE_VERSION) $(IMAGE_PREFIX)postgresql:latest
 
-concent-api: $(CONCENT_API_SOURCE) build/concent-api/requirements.lock build/concent-api/testing.pref
+concent-api: $(CONCENT_API_SOURCE) build/concent-api/requirements.lock build/concent-api/testing.pref build/repositories/concent/
 	docker build --tag $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) build/concent-api/
 	docker tag $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) $(IMAGE_PREFIX)concent-api:latest
 
-verifier: $(VERIFIER_SOURCE) build/verifier/golem/scripts/ build/verifier/golem/imgverifier-requirements.txt
+verifier: $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/golem/scripts/ build/verifier/golem/imgverifier-requirements.txt
 	docker build --tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) build/verifier/
 	docker tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) $(IMAGE_PREFIX)verifier:latest
 
@@ -97,7 +87,7 @@ concent-api-clean:
 verifier-clean:
 	docker rmi $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) --force
 
-build/virtualenv/: $(CONCENT_DIR)/concent_api/requirements.lock build/package-builder/build-virtualenv.sh
+build/virtualenv/: build/repositories/concent/concent_api/requirements.lock build/package-builder/build-virtualenv.sh
 	build/package-builder/build-virtualenv.sh "$<" "build"
 
 build/concent-api/requirements.lock: build/concent-api/concent-api.tar
@@ -111,23 +101,23 @@ build/concent-api/requirements.lock: build/concent-api/concent-api.tar
 build/concent-api/testing.pref: ../concent-builder/files/testing.pref
 	cp "$<" "$@"
 
-build/concent-api/concent-api.tar: $(CONCENT_DIR)  build/virtualenv/
+build/concent-api/concent-api.tar: build/repositories/concent/  build/virtualenv/
 	mkdir --parents $(dir $@)
-	package-builder/build-concent-api.sh "$(CONCENT_DIR)" "build/concent-api"
+	package-builder/build-concent-api.sh "build/repositories/concent/" "build/concent-api"
 
 
-build/nginx-proxy/concent-api-assets.tar: $(CONCENT_DIR) build/virtualenv/ build/package-builder/build-concent-api-assets.sh
+build/nginx-proxy/concent-api-assets.tar: build/repositories/concent/ build/virtualenv/ build/package-builder/build-concent-api-assets.sh
 	mkdir --parents "$(dir $@)"
 	build/package-builder/build-concent-api-assets.sh \
-		"$(CONCENT_DIR)"                              \
+		"build/repositories/concent/"                 \
 		"$(abspath package-builder)"                  \
 		"$(abspath build)"                            \
 		"$(dir $@)"
 
-build/nginx-proxy/concent-api-docs.tar: $(CONCENT_DIR) build/package-builder/build-concent-api-docs.sh
+build/nginx-proxy/concent-api-docs.tar: build/repositories/concent/ build/package-builder/build-concent-api-docs.sh
 	mkdir --parents $(dir $@)
 	build/package-builder/build-concent-api-docs.sh   \
-		"$(CONCENT_DIR)"                              \
+		"build/repositories/concent/"                 \
 		"$(abspath package-builder)"                  \
 		"$(abspath build)"                            \
 		"$(dir $@)"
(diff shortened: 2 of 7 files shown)


concent-builder/repositories.yml
```diff
@@ -3,7 +3,5 @@ repositories:
     url: https://github.com/golemfactory/concent.git
   concent-deployment:
     url: https://github.com/golemfactory/concent-deployment.git
-  concent-deployment-values:
-    url: https://github.com/golemfactory/concent-deployment-values.git
```
cameel: This should be in a separate commit.

containers/Makefile
```diff
@@ -36,15 +36,8 @@ VERIFIER_SOURCE :=                           \
...
-    CONCENT_DIR:= $(abspath ../../concent)
-endif
-$(info CONCENT_DIR=$(CONCENT_DIR))
-
-ifndef GOLEM_DIR
-    GOLEM_DIR:= $(abspath ../../golem)
-endif
-$(info GOLEM_DIR=$(GOLEM_DIR))
+CONCENT_DIR:= build/repositories/concent/
+GOLEM_DIR:= build/repositories/golem/
```
cameel: Please align and fix whitespace here.

containers/versions.yml
```diff
@@ -1,6 +1,8 @@
 debian_version:               "9.4"
 docker_version:               "17.05.0~ce-0~debian-stretch"
 geth_version:                 "v1.8.2"
+golem_version:                "0.13.0"
+concent_version:              "v0.6.0-39-gb882ecc"
```
cameel: So it's possible to use the output of `git describe` with `git checkout`? Cool. I did not expect that.

containers/package-builder/checkout-concent-version.sh.j2
```diff
@@ -0,0 +1,13 @@
...
+DIRECTORY=$1
+
+# Check if a directory exists
+if [ -d "$DIRECTORY" ]; then
+    git fetch
+else
+    git clone /var/concent-data/concent/ "$DIRECTORY"
+    cd "$DIRECTORY"
+    git fetch
+    git checkout {{ concent_version }}
```
cameel: Please make it a parameter of the script.

containers/package-builder/checkout-concent-version.sh.j2
```diff
@@ -0,0 +1,13 @@
...
+
+DIRECTORY=$1
+
+# Check if a directory exists
+if [ -d "$DIRECTORY" ]; then
+    git fetch
+else
+    git clone /var/concent-data/concent/ "$DIRECTORY"
+    cd "$DIRECTORY"
+    git fetch
```
cameel: You do `git fetch` in both cases so you could put it after the loop.

cameel: When running `git fetch`, always specify the remote name. Otherwise the behavior is ambiguous maybe affected by git configuration.

containers/package-builder/checkout-concent-version.sh.j2
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+DIRECTORY=$1
```
cameel: Use lower case characters when naming shell variables.

containers/package-builder/checkout-golem-version.sh.j2
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
```
cameel: You could have a single script that works for both golem and concent if you added more parameters.

containers/package-builder/checkout-concent-version.sh.j2
```diff
@@ -0,0 +1,13 @@
...
+
+# Check if a directory exists
+if [ -d "$DIRECTORY" ]; then
+    git fetch
+else
+    git clone /var/concent-data/concent/ "$DIRECTORY"
```
cameel: Please make a shallow clone. There's no need to clone the whole repository if we only need the last revision. See the parameters that were used in the Ansible playbook.

containers/Makefile
```diff
@@ -135,14 +128,20 @@ $(VERIFIER_SOURCE_FROM_CONCENT_API): build/verifier/%: build/concent-api/%
...
 
-build/verifier/golem/imgverifier-requirements.txt: $(GOLEM_DIR)/apps/rendering/resources/images/requirements.txt
+build/verifier/golem/imgverifier-requirements.txt: $(GOLEM_DIR)apps/rendering/resources/images/requirements.txt
 	cp "$<" "$@"
 
+$(CONCENT_DIR): build/package-builder/checkout-concent-version.sh /var/concent-data/concent
+	"$<" "$@"
+
+$(GOLEM_DIR): build/package-builder/checkout-golem-version.sh /var/concent-data/golem
+	"$<" "$@"
```
cameel: Please hard-code the script name here. It's easier to understand that way.

containers/Makefile
cameel: Please remove these variables. They can no longer be changed by the user who runs make. Leaving them here may imply that they can be changed.

concent-builder/repositories.yml
bartoszbetka: Done.

containers/Makefile

containers/package-builder/checkout-concent-version.sh.j2
bartoszbetka: Could you be more specific ? How do you return the template variable at the script input?

containers/package-builder/checkout-concent-version.sh.j2
bartoszbetka: Done.

bartoszbetka: Done.

containers/package-builder/checkout-concent-version.sh.j2
bartoszbetka: Done.

containers/package-builder/checkout-concent-version.sh.j2
cameel: I mean that instead of this you could have something like
``` bash
version=$2
git checkout $version
```
and call the script like
``` bash
./checkout-concent-version.sh <version>
```

Oh, but I can see the problem now. Version is embedded in `version.yml` and you don't have it directly in the makefile. Well, in that case let's leave it as is.

containers/package-builder/checkout-golem-version.sh.j2
bartoszbetka: Done.

containers/package-builder/checkout-concent-version.sh.j2
bartoszbetka: There is no parameter `--depth` in ansible playbook. This not work, because we clone repository locally.

containers/Makefile

containers/package-builder/checkout-concent-version.sh.j2
bartoszbetka: Ok.

containers/package-builder/checkout-repositories-version.sh.j2
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
+
+directory=$1
+repository=$2
```
cameel: Please reverse the order of these parameters. The script is basically a wrapper over `git clone` and the order in clone is `<repository> <directory>`. It will make it easier to remember.

containers/package-builder/checkout-repositories-version.sh.j2
```diff
@@ -0,0 +1,26 @@
...
+        git clone /var/concent-data/golem/ "$directory"
+        cd "$directory"
+        git checkout {{ golem_version }}
+fi
+
+if [[ "$repository" == "concent" ]]; then
+git fetch https://github.com/golemfactory/concent.git
+elif [[ "$repository" == "golem" ]]; then
+git fetch https://github.com/golemfactory/golem.git
+else
```
cameel: Please fix indentation here.

containers/package-builder/checkout-repositories-version.sh.j2
```diff
@@ -0,0 +1,26 @@
...
+if [ ! -d "$directory" ] && [[ "$repository" == "golem" ]]; then
+        git clone /var/concent-data/golem/ "$directory"
+        cd "$directory"
+        git checkout {{ golem_version }}
+fi
+
+if [[ "$repository" == "concent" ]]; then
+git fetch https://github.com/golemfactory/concent.git
+elif [[ "$repository" == "golem" ]]; then
+git fetch https://github.com/golemfactory/golem.git
```
cameel: Don't fetch from github. You should fetch from the repository in `/var/`.

containers/package-builder/checkout-repositories-version.sh.j2
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
+
+directory=$1
+repository=$2
+
+# Check if a directory exists
+if [ ! -d "$directory" ] && [[ "$repository" == "concent" ]]; then
```
cameel: `repository` should be `/var/concent-data/concent/`, not `/concent/`. Similarly for `golem`. This way you won't have to hard-code it for `concent` and `golem`.

containers/Makefile
```diff
@@ -135,14 +125,20 @@ $(VERIFIER_SOURCE_FROM_CONCENT_API): build/verifier/%: build/concent-api/%
...
 
-build/verifier/golem/imgverifier-requirements.txt: $(GOLEM_DIR)/apps/rendering/resources/images/requirements.txt
+build/verifier/golem/imgverifier-requirements.txt: build/repositories/golem/apps/rendering/resources/images/requirements.txt
 	cp "$<" "$@"
 
+build/repositories/concent/: build/package-builder/checkout-repositories-version.sh /var/concent-data/concent
+	build/package-builder/checkout-repositories-version.sh "$@" "concent"
+
+build/repositories/golem/: build/package-builder/checkout-repositories-version.sh /var/concent-data/golem
+	build/package-builder/checkout-repositories-version.sh "$@" "golem"
```
cameel: Please rename the script to `clone-or-update.sh` or maybe `get-code.sh`.

containers/package-builder/checkout-repositories-version.sh.j2
bartoszbetka: Done.

containers/package-builder/checkout-repositories-version.sh.j2
bartoszbetka: Done.

containers/package-builder/checkout-repositories-version.sh.j2
bartoszbetka: Done.

containers/package-builder/checkout-repositories-version.sh.j2
bartoszbetka: Done.

containers/package-builder/clone-or-update.sh.j2
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+
+repository=$1
+directory=$2
+
+# Check if a directory exists
+if [ ! -d "$directory" ] && [[ "$repository" == "/var/concent-data/concent" ]]; then
+    git clone "$repository" "$directory"
+    cd "$directory"
+    git checkout {{ concent_version }}
```
cameel: The code in both `if`s is now identical. You can replace it with a single `if`.

bartoszbetka: It's not possible to do that. How would you recognize which template variable use ?

containers/package-builder/clone-or-update.sh.j2
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -e
+
+repository=$1
+directory=$2
+
+if [[ "$repository" == "concent" ]]; then
+    repository_path=/var/concent-data/concent/
+    repository_version={{ concent_version }}
```
cameel: Add quotes around `{{ }}` and it's good.

PR #142, opened 2018-05-29
Add feature that add specific version to config maps and secrets

### LINKED ISSUE #135: Versioning config maps and secrets
To be able to run a rolling update on Kubernetes our pods need to be able to access both the old and the new version of each config map and secret. Kubernetes does not take care of it automatically so we have to do versioning manually.

Modify our deployment system so that:
- Deployment version is a concatenation of versions of `concent-deployment` and `concent-deployment-values`.
    - This version should be available in a variable called `DEPLOYMENT_VERSION` in the Makefile. Then added as a context variable when rendering templates.
    - To get the version of `concent-deployment-values` you'll need to run `git describe` in the Ansible playbook on the local machine since that's where we take `var.yml` from.
- Each secret and config map name should contain this version as a suffix.
    - For the ones created directly with a `kubectl` call, just add it to the name.
    - For the ones that have separate `.yml` files, use it in the name inside the file but you can leave the name of the `.yml` file itself without the suffix.
- Use the suffix in every `.yml` file that references the config map or secret.
- Note that after this change `delete-config-maps.sh` will no longer be able to delete the old config maps from previous deployments. Same with `delete-job-dependencies.sh`. We will solve that long-term by keeping the old deployment scripts but in the short term please work around it by making both of these scripts list all the config maps present on the cluster, finding those with the matching name (ignoring the suffix) and deleting them all, no matter whether the version matches or not.

bartoszbetka:
Resolves #135

### DIFF
--- a/kubernetes/services/verifier.yml.j2
+++ b/kubernetes/services/verifier.yml.j2
@@ -38,9 +38,6 @@ spec:
         - name: concent-api-secrets-{{ deployment_version }}
           secret:
             secretName: concent-api-secrets-{{ deployment_version }}
-        - name: concent-api-settings-{{ deployment_version }}
-          configMap:
-            name: concent-api-settings-{{ deployment_version }}
         - name: verifier-settings-{{ deployment_version }}
           configMap:
             name: verifier-settings-{{ deployment_version }}


cameel:
We're going to start using Helm/Tiller in the near future and it looks like it already has a feature like this. See issue 4044 in https://github.com/helm/helm. This means that there's no point in doing this on our side and complicating our scripts after all. We probably won't be merging this pull request but I'm leaving it open in case anything changes.

PR #145, opened 2018-06-05
Add `ignore_errors` directive to `Stop postgres container` task

bartoszbetka:
The ansible `always` directive do not perform second task if first one
occurs errors. We want to be sure that all tasks are always executed.

### DIFF
--- a/concent-builder/build-test-and-push.yml
+++ b/concent-builder/build-test-and-push.yml
@@ -85,6 +85,7 @@
       always:
         - name:  Stop postgres container
           shell: docker stop {{ cluster }}-{{ image_prefix }}postgresql-unittest
+          ignore_errors: yes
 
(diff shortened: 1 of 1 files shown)


concent-builder/build-test-and-push.yml
```diff
@@ -85,6 +85,7 @@
       always:
         - name:  Stop postgres container
           shell: docker stop {{ cluster }}-{{ image_prefix }}postgresql-unittest
+          ignore_errors: yes
```
cameel: If it's needed here, don't we need it for `docker network rm` as well?

bartoszbetka: If there was another task after `docker network rm`, we would also need `ignore_errors` to perform it.

PR #153, opened 2018-06-08
Change old not needed setting to assertion

bartoszbetka:
This setting is not needed after default database name was set for empty
in concent side.


cameel:
Please make the commit message more specific.

PR #160, opened 2018-06-21
Feature add installation of concent in the virtual machine

### LINKED ISSUE #147: Automatic installation of Concent in the virtual machine
This issue expands on #146. The goal is to create an Ansible playbook that automates the installation steps described in [Concent README](https://github.com/golemfactory/concent/blob/master/README.md).

1. Create an additional Ansible playbook in `concent-vm/` that can be run against the machine.
2. The playbook should expect a file called `extra_settings.py` somewhere within its directory
    - Add it to `.gitignore`. It must be provided by the user.
    - The file will be used to define `CONCENT_ETHEREUM_PRIVATE_KEY`, `CONCENT_ETHEREUM_ADDRESS` and other mandatory settings not defined in `development.py`. It may also be used to pass any other extra settings to the Concent instance running in the machine.
3. The playbook should perform the following steps:
     - Clone `concent` repository and create a working copy in the home directory.
        - Fetch the latest code if the repository already exists.
        - Check out the `master` branch. User should be able to specify a different branch using Ansible's `--extra-vars`.
    - Create a virtualenv for Concent and install packages listed in`concent_api/requirements.lock` and `requirements-development.txt`.
    - Install gunicorn in the virtualenv.
    - Upload `extra_settings.py` to the `settings/` directory in the working copy.
    - Create `local_settings.py` for Concent from a template:
        - Import settings from `development.py`.
        - Import settings from `extra_settings.py`. This should be the last instruction in the file.
        - Configure the database, Geth, PostgreSQL and storage-server-related settings. Assume these services are all running on `localhost`.
    - Migrate the databases.
    - Create a superuser account in Django if one does not exist yet.
    - Run `full_check.sh`.

bartoszbetka:
Resolves #147


concent-vm/install-concent.yml
```diff
@@ -0,0 +1,90 @@
...
+              app_path:   "{{ concent_dir }}/concent_api"
+              virtualenv: "~/virtualenv"
+          with_items:
+            - control
+            - storage
+
+        - name:   Check if superuser already exist
+          shell: echo "from django.contrib.auth import get_user_model; User = get_user_model(); print(User.objects.filter(username='admin').count()>0)" | ~/virtualenv/bin/python ./manage.py shell
+          args:
+            chdir: "{{ concent_dir }}/concent_api"
```
cameel: Please make it a proper script. This is a bit hacky.

Call it `user-exists.py`.

You'll be able to execute it with `python` (instead of piping to `manage.py shell`) if you start the script with:

``` python
import os
import django
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "concent_api.settings")
django.setup()
```
and put it in the same directory as `manage.py`.

You could also set the `DJANGO_SETTINGS_MODULE` variable from ansible:
``` yaml
  name: <task name>
  ...
  environment:
    DJANGO_SETTINGS_MODULE: concent_api.settings
```

concent-vm/install-concent.yml
```diff
@@ -0,0 +1,90 @@
...
+    - vars.yml
+  tasks:
+    - become:      yes
+      become_user: vagrant
+      block:
+        - name:   Clone concent repository
+          git:
+            repo:    "{{ repositories.concent.url }}"
+            dest:    "{{ concent_dir }}"
+            version: "{{ concent_version }}"
```
cameel: The issue states that this should check out code from `master` by default. As it is now, it'll use version from `versions.yml`.

It's fine but you should have mentioned this change in a comment in the issue.

concent-vm/install-concent.yml
```diff
@@ -0,0 +1,90 @@
+- hosts:
+    - vagrant
+  vars_files:
+    - ../containers/versions.yml
+    - ../concent-builder/repositories.yml
+    - vars.yml
+  tasks:
```
michalpiasecki1: wouldnt be better to put ansible logic in separate roles? It is covered in best practicies: https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html 

"One crucial way to organize your playbook content is Ansible’s “roles” organization feature, which is documented as part of the main playbooks page. You should take the time to read and understand the roles documentation"

Ideally, playbook should be just mapping between role and a host. Each role can have different set of vars, files and templates. This modular design is easier to maintain, reuse and understand.

Roles can include:
-PosgreSQL (install and migrate data)
-nginx (install and setup service)
-prepare directories and users
-system level dependencies
-build and run concent
-run tests

cameel: I fully agree. I have even suggested splitting it in the pull request this one builds upon (https://github.com/golemfactory/concent-deployment/pull/159#discussion_r204779817). It was less of a problem when we had only a couple playbooks but now they're starting to get repetitive so it's a good idea to split them into smaller parts.

I suggested just putting tasks into smaller playbooks and importing them to keep things simple for now. But roles are fine too.

@bartoszbetka: @michalpiasecki1 has lots of experience with Ansible so please talk with him if you have any questions about how to split it.

BTW, @michalpiasecki1 if you want to create a bullet list in Markdown you have to put a space between `-` and the text. I have edited your comment to show that. See how it looks now.

concent-vm/install-concent.yml
```diff
@@ -0,0 +1,99 @@
...
+        - name:   Copy extra_settings.py from host computer to the remote server
+          copy:
+            src:    "{{ item }}"
+            dest:   "{{ concent_dir }}/concent_api/concent_api/settings/{{ item }}"
+            owner:  vagrant
+            group:  vagrant
+            mode:   0644
+          with_items:
+            - extra_settings.py
+            - local_settings.py
```
cameel: Only `extra_settings.py` was supposed to be copied to the machine.
`local_settings.py` should be created from a template.

concent-vm/install-concent.yml
```diff
@@ -0,0 +1,99 @@
...
+        - name:   Clone concent repository
+          git:
+            repo:    "{{ repositories.concent.url }}"
+            dest:    "{{ concent_dir }}"
+            version: "{{ concent_version }}"
+            clone:   yes
+            update:  yes
+
+        - name:   Create virtualenv for concent
+          command:  virtualenv ~/virtualenv --python python3.6 creates="~/virtualenv"
```
cameel: You should use `python_version` from `versions.yml`.

concent-vm/install-concent.yml
cameel: Oh, nevermind. I see that this file is actually stored in the repository.

The task `name` is just confusing because it suggests that this part is supposed to upload user's own settings. I think these two uploads should be in separate steps with descriptions that make things clear.

PR #161, opened 2018-06-21
Change log level from `info` to `debug` in django apps

bartoszbetka:


containers/concent-api/run-gunicorn.sh.j2
```diff
@@ -7,5 +7,5 @@ exec "/srv/http/virtualenv/bin/gunicorn" concent_api.wsgi:application  \
     --workers {{ gunicorn_workers }}       \
     --timeout {{ gunicorn_timeout }}       \
     --bind=0.0.0.0:80                      \
-    --log-level=info                       \
+    --log-level=debug                      \
```
cameel: This is not enough. You must also change the log level in the `console` handler in Concent:
https://github.com/golemfactory/concent/blob/master/concent_api/concent_api/settings/base.py#L168-L172

But this requires a separate request in that repo so I'm accepting this one.

PR #165, opened 2018-06-25
Add missing `CELERY_BROKER_URL` setting to conductor local_settings.py file

bartoszbetka:
The `CELERY_BROKER_URL` setting contain address to rabbitmq. Without
this setting, the conductor could not connect to rabbitmq.
Resolves [#499](https://github.com/golemfactory/concent/issues/499)


cameel:
What about other workers? Do they all have this setting already?

bartoszbetka:
The workers always had to this setting. We forgot add it to `conductor-url`.

PR #166, opened 2018-06-25
Change and add new time settings in `timing_settings.py` file to meet new requirements

bartoszbetka:


kbeker:
We want these particular times settings because we don't want to wait a lot of time in api-tests for correct response from Concent :)

cameel:
But why not a second shorter? Or a second longer? Why these particular values?

kubernetes/config-maps/concent/timing_settings.py.j2
```diff
@@ -1,7 +1,11 @@
 {% if fast_timing %}
-ADDITIONAL_VERIFICATION_CALL_TIME = 15
+CUSTOM_PROTOCOL_TIMES = True
```
kbeker: We needed to add this, because this was hardcoded in base.py and we couldn't change it during tests.

kbeker:
@cameel This times are copied from settings from `testing.py`

PR #167, opened 2018-06-26
Move `VERIFIER_MIN_SSIM` setting from `base.py` file in concent repository to verifier local_settings.py file

bartoszbetka:
This setting is moved, because it only needed when `CONCENT_FEATURES`
contains verifier app. Second problem is that system check occurs error
when `VERIFIER_MIN_SSIM` is define without enable verifier in
`CONCENT_FEATURES`.

It's depend on [#558](https://github.com/golemfactory/concent/pull/558)


cameel:
@bartoszbetka When closing a pull request please remember to always explain why you're closing it.

PR #175, opened 2018-07-03
Bump geth version up to v1.8.11

### LINKED ISSUE #107: Update Geth to the latest version
Right now the latest version is 1.8.3. We're still on 1.8.2 because there was a change in a Dockerfile that affects file permissions (geth no longer runs as `root`). The result is that the blockchain files on the persistent disk are no longer accessible to the user in the new image because they're still owned by root.

Fix it on the dev and staging clusters by manually changing permissions. It's fine to just clear the persistent disk if that's easier since these are not production clusters.

bartoszbetka:
Resolves #107

### DIFF
--- a/containers/versions.yml
+++ b/containers/versions.yml
@@ -1,6 +1,6 @@
 debian_version:               "9.4"
 docker_version:               "17.05.0~ce-0~debian-stretch"
-geth_version:                 "v1.8.2"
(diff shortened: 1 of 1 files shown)


bartoszbetka:
Should be delivered after [#605](https://github.com/golemfactory/concent/issues/605)

PR #180, opened 2018-07-04
Update openresty version to the newest one

### LINKED ISSUE #178: Update openresty version
Update openresty version requires change in docker images. They need some additionial lib or it's not compatible with `nginx-big-upload` module.

bartoszbetka:
Resolves #178


containers/nginx-storage/install-dependencies.sh.j2
```diff
@@ -15,7 +15,9 @@ opm get spacewander/luafilesystem
 # Create .so symlink for libcrypto.so manually.
 # In normal circumstances the link would be created by ldconfig,
 # but in Alpine Linux ldconfig is non-functional.
-ln -s /usr/lib/libcrypto.so.41 /usr/lib/libcrypto.so
+# WARNING !!! If you update openresty version, remember to check
+# that `libcrypto.so.XX` wasn't changed.
+ln -s /usr/lib/libcrypto.so.42 /usr/lib/libcrypto.so
```
cameel: Wouldn't it be better to say something like `"if you get error "<error message here>", you need to update the version number in libcrypto.so.XX"`? As it is now it seems to imply that you should not update it.

bartoszbetka: Done

containers/nginx-storage/install-dependencies.sh.j2
```diff
@@ -15,7 +15,9 @@ opm get spacewander/luafilesystem
 # Create .so symlink for libcrypto.so manually.
 # In normal circumstances the link would be created by ldconfig,
 # but in Alpine Linux ldconfig is non-functional.
-ln -s /usr/lib/libcrypto.so.41 /usr/lib/libcrypto.so
+# WARNING !!! if you get error "<Error loading shared library libcrypto.so: No such file or directory>",
+# you need to update the version number in libcrypto.so.XX
```
cameel: The `<>` was only to show that it's a place holder :) I did not mean to include it in the message :)

PR #181, opened 2018-07-06
Add `MOCK_VERIFICATION_ENABLED` setting to verifier `local_settings.py` file to enable or disable mocked verification

bartoszbetka:


kubernetes/config-maps/verifier/local_settings.py.j2
```diff
@@ -32,6 +32,7 @@ CONCENT_FEATURES = ["verifier"]
 
 STORAGE_SERVER_INTERNAL_ADDRESS = "http://nginx-storage.default.svc.cluster.local/"
 VERIFIER_STORAGE_PATH = "/srv/storage/"
+MOCK_VERIFICATION_ENABLED = {{ mock_verification_enabled }}
```
cameel: Please make sure that the value that ends up in the file after template rendering is actually `True` or `False` and not a string.

bartoszbetka: Done.

PR #189, opened 2018-07-18
Add signing service container to build system

### LINKED ISSUE #185: Signing Service container
This is a part of the implementation of the "Concent Signing Service" blueprint (https://github.com/golemfactory/concent/issues/597).

#### Overview
We need to have a separate container for the Signing Service. It should be able to run both on Kubernetes and without it. We want to be able to set it up ourselves on test clusters but in production it will be running on Golem's infrastructure.

In addition to building the container our build system should also be able to generate a self-contained package that allows anyone to build the service.

#### Building the container
The container should be based on Debian or Alpine Linux.

1. Dependencies
    - Install system dependencies.
    - Install Python dependencies from `concent/signing_service/requirements.lock`.
        - This is not a Django project so it does not require anything extra to serve it (gunicorn, celery, etc.).
2. Signing Service
    - Install source from `concent/signing_service/` in the container.
        - It's very likely that the service will require some files from the other directories. We'll know for sure when the service is implemented. Be prepared to update the container if that turns out to be the case.
    - Include a file with version
        - Signing Service uses has the same version as Concent.
    - When building, put the source of the service in a subdirectory of the directory that contains the `Dockerfile`. The directory should contain everything that's needed to build it.
3. Runner
    - Create a simple runner script that will execute the script from `concent/signing_service/` and pass all command-line parameters to it. Put it in `/usr/local/bin` so that we don't have to use the full path to run it.
    - Include an entry point that runs the script as non-root user.
    - The container should run the runner script by default when it starts .

#### Source package
Add a step in `containers/Makefile` that creates `concent-signing-service-<version>.tar.xz` archive that contains:
- Dockerfile (after template rendering) and all the scripts it installs in the machine
- The source of the Signing Service
- Version file

Create a phony target called `concent-signing-service-package` that can be used to build it without specifying the version on the command line.

It should be possible to build the container simply by unpacking the package and running `docker build` inside.

#### Instructions
In README include instructions for both building the package and manually building the container by extracting the package. Include a link to the `Dockerfile` in the repository so that someone reading the docs can easily take a look at it.

Also show how to run the container and how to attach a volume with Ethereum public key and pass the full path to that file on the command line.

bartoszbetka:
Resolves #185 
It's depend on [#631](https://github.com/golemfactory/concent/issues/631),  [#632](https://github.com/golemfactory/concent/issues/632), [#633](https://github.com/golemfactory/concent/issues/633)

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,3 +241,90 @@ docker run                         \
 curl http://localhost:8001/
 ```
 
+## Concent Signing Service
+In addition to Concent itself, this repository contains files necessary to build the Concent Signing Service.
+`Makefile` builds a Docker container but also produces a source package that includes `Dockerfile` and all source files needed to build it.
+The package can be used to build the container without having to set up `concent-deployment`.
+
+### Building the Signing Service package
+This is only needed if you want to build the package yourself.
+If you have received the package and only want to build and run the docker container, you can skip this section.
+
+1. Install dependencies needed to build containers and render configuration files from templates.
+
+    Example for Ubuntu:
+
+    ``` bash
+    apt-get update
+    apt-get install make python3 python-pip python3-yaml
+    apt-get install docker.io
+    pip install yasha
+    ```
+
+2. Ensure that you can start docker containers.
+    It's recommended to add your user to the `docker` group (make sure that this group exists) so that you don't have to do this as root.
+    Note that the change will not take effect until you log out of the current shell session.
+
+    Example for Ubuntu:
+
+    ``` bash
+    sudo groupadd docker
+    sudo usermod --all --groups docker <your user name>
+    ```
+
+3. Run make
+
+    ```bash
+    cd containers/
+    make concent-signing-service-package
+    ```
+
+### Building the Signing Service from the package
+1. Extract the package
+2. Go to `signing_service/` directory inside the package and build the Docker image
+
+    ``` bash
+    cd signing_service/
+    docker build --tag concent-signing-service:$(cat signing-service/RELEASE-VERSION) .
+    docker tag                                                         \
+        concent-signing-service:$(cat signing-service/RELEASE-VERSION) \
+        concent-signing-service:latest
+    ```
+
+### Running the Signing Service
+To run it in a docker container with access to your local network interface, run:
+
+```bash
+docker run                                                  \
+    --detach                                                \
+    --env         ETHEREUM_PRIVATE_KEY                      \
+    --env         SIGNING_SERVICE_PRIVATE_KEY               \
+    --env         SENTRY_DSN                                \
+    --network     host                                      \
+    --hostname    signing-service                           \
+    --name        signing-service                           \
+    --restart     on-failure                                \
+    concent-signing-service                                 \
+        concent.golem.network                               \
+        85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us    \
+        --concent-cluster-port                  9055        \
+        --ethereum-private-key-from-env                     \
+        --signing-service-private-key-from-env              \
+        --sentry-dsn-from-env                               \
+        --sentry-environment                    mainnet
+```
+
+This assumes that:
+- The service can connect to a Concent cluster at `concent.golem.network:9055`.
+- `85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us` is Concent's public key encoded in base64.
+- There's a shell variable called `ETHEREUM_RPIVATE_KEY` and it contains base64-encoded private key of the Concent contract.
+- There's a shell variable called `SIGNING_SERVICE_PRIVATE_KEY` and it contains base64-encoded private key for signing Golem Messages created by the Signing Service.
+- There's a shell variable called `SENTRY_DSN` that contains the secret ID that allows submitting crash report to a [Sentry](https://sentry.io) project.
+    The one given above is just an example.
+    If so, you need to provide a valid DSN for the project that should receive the reports.
+    Otherwise just skip the `--sentry-dsn-from-env` parameter and the `SENTRY_DSN` variable.
+- `mainnet` is the name of the Sentry environment that should be included in error reports.
+
+Note that the service will crash on errors.
+The host system is responsible for restarting it in that case.
+If it's running in a Docker container you can easily achieve this with the `--restart on-failure` option.
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -46,6 +46,16 @@ VERIFIER_SOURCE :=                           \
     $(VERIFIER_SOURCE_FROM_CONCENT_API)
 
 
+SIGNING_SERVICE_SOURCE :=                               \
+    build/signing-service/Dockerfile                    \
+    build/signing-service/install-dependencies.sh       \
+    build/signing-service/install-signing-service.sh    \
+    build/signing-service/run-signing-service.sh        \
+    build/signing-service/clean-container.sh            \
+    build/signing-service/create-user.sh                \
+    build/signing-service/build-dependencies.txt        \
+    build/signing-service/testing.pref
+
 ifndef IMAGE_VERSION
     IMAGE_VERSION := $(shell git describe --always --abbrev=16)
 endif
@@ -57,7 +67,7 @@ endif
 $(info IMAGE_PREFIX=$(IMAGE_PREFIX))
 
 
-all: nginx-proxy nginx-storage concent-api postgresql verifier
+all: nginx-proxy nginx-storage concent-api postgresql verifier signing-service
 
 nginx-proxy: build/ $(NGINX_PROXY_SOURCE) build/nginx-proxy/concent-api-assets.tar build/nginx-proxy/concent-api-docs.tar
 	docker build --tag $(IMAGE_PREFIX)nginx-proxy:$(IMAGE_VERSION) build/nginx-proxy/
@@ -79,7 +89,11 @@ verifier: build/ $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/gol
 	docker build --tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) build/verifier/
 	docker tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) $(IMAGE_PREFIX)verifier:latest
 
-docker-clean: nginx-proxy-clean concent-api-clean postgresql-clean verifier-clean
+signing-service: build/ $(SIGNING_SERVICE_SOURCE) build/signing-service/signing-service/ build/signing-service/middleman_protocol/ build/signing-service/signing-service/RELEASE-VERSION
+	docker build --tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) build/signing-service/
+	docker tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) $(IMAGE_PREFIX)signing-service:latest
+
+docker-clean: nginx-proxy-clean concent-api-clean postgresql-clean verifier-clean signing-service-clean
 
 nginx-proxy-clean:
 	docker rmi $(IMAGE_PREFIX)nginx-proxy:$(IMAGE_VERSION) --force
@@ -96,6 +110,9 @@ concent-api-clean:
 verifier-clean:
 	docker rmi $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) --force
 
+signing-service-clean:
+	docker rmi $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) --force
+
 build/virtualenv/: build/repositories/concent/concent_api/requirements.lock build/package-builder/build-virtualenv.sh
 	build/package-builder/build-virtualenv.sh "$<" "build"
 
@@ -107,7 +124,7 @@ build/concent-api/requirements.lock: build/concent-api/concent-api.tar
 		--directory build/concent-api/           \
 		concent_api/requirements.lock
 
-build/concent-api/testing.pref: ../concent-builder/files/testing.pref
+build/%/testing.pref: ../concent-builder/files/testing.pref
 	cp "$<" "$@"
 
 build/concent-api/concent-api.tar: build/repositories/concent  build/virtualenv/
@@ -155,6 +172,15 @@ build/repositories/golem/: build/package-builder/clone-or-update.sh /var/concent
 build/%/middleman_protocol/: build/repositories/concent
 	cp --recursive "$</middleman_protocol/." "$@"
 
+build/signing-service/signing-service/RELEASE-VERSION: build/signing-service/signing-service/
+	echo -n "$(shell cd build/repositories/concent/; git describe --always --abbrev=16)" > "$@"
+
+build/signing-service/signing-service/: build/repositories/concent
+	cp --recursive "$</signing_service/." "$@"
+
+build/signing-service/middleman_protocol/: build/repositories/concent
+	cp --recursive "$</middleman_protocol/." "$@"
+
 build/:
 	mkdir --parents "$@"
 
@@ -177,15 +203,16 @@ build/config-with-versions.yml: versions.yml config.yml ../tools/merge-yaml.py
 
 
 clean:
-	rm -rf build/
+	rm -rf build/ dist/
 
 
-.PHONY:                                 \
-    all                                 \
-    nginx-proxy   nginx-proxy-clean     \
-    nginx-storage nginx-storage-clean   \
-    postgresql    postgresql-clean      \
-    concent-api   concent-api-clean     \
-    verifier      verifier-clean        \
-    docker-clean                        \
+.PHONY:                                   \
+    all                                   \
+    nginx-proxy     nginx-proxy-clean     \
+    nginx-storage   nginx-storage-clean   \
+    postgresql      postgresql-clean      \
+    concent-api     concent-api-clean     \
+    verifier        verifier-clean        \
+    signing-service signing-service-clean \
+    docker-clean                          \
     clean
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -67,7 +67,7 @@ endif
 $(info IMAGE_PREFIX=$(IMAGE_PREFIX))
 
 
-all: nginx-proxy nginx-storage concent-api postgresql verifier signing-service
+all: nginx-proxy nginx-storage concent-api postgresql verifier signing-service concent-signing-service-package
 
 nginx-proxy: build/ $(NGINX_PROXY_SOURCE) build/nginx-proxy/concent-api-assets.tar build/nginx-proxy/concent-api-docs.tar
 	docker build --tag $(IMAGE_PREFIX)nginx-proxy:$(IMAGE_VERSION) build/nginx-proxy/
@@ -92,6 +92,18 @@ verifier: build/ $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/gol
 signing-service: build/ $(SIGNING_SERVICE_SOURCE) build/signing-service/signing-service/ build/signing-service/middleman_protocol/ build/signing-service/signing-service/RELEASE-VERSION
 	docker build --tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) build/signing-service/
 	docker tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) $(IMAGE_PREFIX)signing-service:latest
+
+concent-signing-service-package: build/ $(SIGNING_SERVICE_SOURCE) dist/concent-signing-service-%.tar.xz
+
+dist/concent-signing-service-%.tar.xz: build/signing-service/signing-service/ build/signing-service/middleman_protocol/ build/signing-service/signing-service/RELEASE-VERSION
+	mkdir dist || true
+	tar                                                                                                                    \
+		--create                                                                                                           \
+		--verbose                                                                                                          \
+		--xz                                                                                                               \
+		--file="dist/concent-signing-service-$(shell cat build/signing-service/signing-service/RELEASE-VERSION).tar.xz"    \
+		--directory="build/"                                                                                               \
+		signing-service/
 
 docker-clean: nginx-proxy-clean concent-api-clean postgresql-clean verifier-clean signing-service-clean
 
@@ -203,7 +215,10 @@ build/config-with-versions.yml: versions.yml config.yml ../tools/merge-yaml.py
 
 
 clean:
-	rm -rf build/ dist/
+	rm -rf build/
+
+distclean:
+	rm -rf dist/
 
 
 .PHONY:                                   \
@@ -214,5 +229,7 @@ clean:
     concent-api     concent-api-clean     \
     verifier        verifier-clean        \
     signing-service signing-service-clean \
+    concent-signing-service-package       \
     docker-clean                          \
+    distclean                             \
     clean
--- a/containers/signing-service/Dockerfile.j2
+++ b/containers/signing-service/Dockerfile.j2
@@ -1,0 +1,21 @@
+FROM debian:{{ debian_version }}
+
+COPY create-user.sh                     /tmp/
+RUN  /tmp/create-user.sh
+
+COPY signing-service/RELEASE-VERSION    /usr/lib/signing_service/RELEASE-VERSION
+COPY signing-service/requirements.lock  /usr/lib/signing_service/signing-service/
+COPY testing.pref                       /etc/apt/preferences.d/
+COPY build-dependencies.txt             /tmp/
+COPY install-dependencies.sh            /tmp/
+COPY install-signing-service.sh         /tmp/
+COPY clean-container.sh                 /tmp/
+COPY middleman_protocol/                /usr/lib/signing_service/middleman_protocol/
+COPY signing-service/RELEASE-VERSION    /usr/lib/signing_service/middleman_protocol/RELEASE-VERSION
+RUN  /tmp/install-signing-service.sh
+
+USER signing-service
+COPY signing-service/                   /usr/lib/signing_service/signing-service/
+COPY run-signing-service.sh             /usr/local/bin/
+
+ENTRYPOINT ["/usr/local/bin/run-signing-service.sh"]
--- a/containers/signing-service/clean-container.sh
+++ b/containers/signing-service/clean-container.sh
@@ -1,0 +1,15 @@
+#!/bin/bash -e
+
+
+readarray temporary_dependencies < ${BASH_SOURCE%/*}/build-dependencies.txt
+
+# Clean up
+apt-get clean
+apt autoremove --assume-yes
+apt-get remove --purge --assume-yes ${temporary_dependencies[*]}
+
+rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Temporary file which is needed to build middleman protocol
+rm /usr/lib/signing_service/RELEASE-VERSION
+
--- a/containers/signing-service/install-dependencies.sh.j2
+++ b/containers/signing-service/install-dependencies.sh.j2
@@ -1,0 +1,27 @@
+#!/bin/bash -e
+
+# Update package lists for Debian and install dependencies
+# Packages needed to build python extensions, dependencies of golem-messages in particular
+readarray dependencies < ${BASH_SOURCE%/*}/build-dependencies.txt
+
+testing_dependencies=(
+    # There's no stable Python 3.6 package in Debian Stretch yet so we have to get it from testing.
+    python{{ python_version }}
+    python{{ python_version }}-dev
+    virtualenv
+)
+
+apt-get update --assume-yes
+apt-get install              \
+    --assume-yes             \
+    --no-install-recommends  \
+    ${dependencies[*]}
+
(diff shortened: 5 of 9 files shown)


containers/Makefile
```diff
@@ -79,7 +89,22 @@ verifier: build/ $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/gol
...
+	docker tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) $(IMAGE_PREFIX)signing-service:latest
+
+concent-signing-service-package: build/ $(CONCENT_API_SOURCE) $(SIGNING_SERVICE_SOURCE) build/concent-signing-service-%.tar.xz
+
+build/concent-signing-service-%.tar.xz:  build/signing-service/signing-service/
+	tar                                                                                                         \
+		--create                                                                                                \
+		--verbose                                                                                               \
+		--xz                                                                                                    \
+		--file="build/concent-signing-service-$(shell cat build/repositories/concent/RELEASE-VERSION).tar.xz"   \
```
cameel: You should have `RELEASE-VERSION` included inside `signing-service/` too. It would be much better to use it here instead of reaching out to concent container's directory.

cameel: `build/repositories/concent/RELEASE-VERSION` should be on the prerequisite list of this target if you want to use it.

containers/Makefile
```diff
@@ -79,7 +89,22 @@ verifier: build/ $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/gol
...
 	docker tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) $(IMAGE_PREFIX)verifier:latest
 
-docker-clean: nginx-proxy-clean concent-api-clean postgresql-clean verifier-clean
+signing-service: build/ $(CONCENT_API_SOURCE) $(SIGNING_SERVICE_SOURCE) build/signing-service/signing-service/
+	docker build --tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) build/signing-service/
+	docker tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) $(IMAGE_PREFIX)signing-service:latest
+
+concent-signing-service-package: build/ $(CONCENT_API_SOURCE) $(SIGNING_SERVICE_SOURCE) build/concent-signing-service-%.tar.xz
+
+build/concent-signing-service-%.tar.xz:  build/signing-service/signing-service/
```
cameel: This should be somewhere where it will be apparent that it's a final product, not an intermediate file. I'd create a `dist/` directory next to `build/`.

containers/Makefile
```diff
@@ -152,6 +180,15 @@ build/repositories/concent/: build/package-builder/clone-or-update.sh /var/conce
 build/repositories/golem/: build/package-builder/clone-or-update.sh /var/concent-data/golem
 	build/package-builder/clone-or-update.sh  "golem" "$@"
 
+$(SIGNING_SERVICE_SOURCE_FROM_CONCENT_API): build/signing-service/%: build/concent-api/%
+	mkdir --parents "build/verifier/"
+	cp "$<" "$@"
```
cameel: Why `verifier`? Looks like a mistake.

containers/Makefile
```diff
@@ -152,6 +180,15 @@ build/repositories/concent/: build/package-builder/clone-or-update.sh /var/conce
...
+
+build/signing-service/signing-service/: build/repositories/concent/ build/concent-api/concent-api.tar
+	cp --recursive "$<signing_service/." "$@"
+	cp "$<RELEASE-VERSION" "$@"
```
cameel: `RELEASE-VERSION` should have its own target.

cameel: Please remove the `/` from the end of the first prerequisite so that in commands you can use `$</RELEASE-VERSION`. Otherwise the `$<` variable looks pretty confusing.

cameel: This target does not look like it really depends on `concent-api.tar`.

If that's because building `concent-api.tar` causes the creation of `RELEASE-VERSION`, you should add a comment that explains it.

containers/signing-service/Dockerfile.j2
```diff
@@ -0,0 +1,15 @@
...
+COPY testing.pref                       /etc/apt/preferences.d/
+COPY install-dependencies.sh            /tmp/
+RUN  /tmp/install-dependencies.sh
+
+COPY signing-service/requirements.lock  /srv/http/signing_service/
+COPY install-signing-service.sh         /tmp/
+RUN  /tmp/install-signing-service.sh
+
+USER signingserviceuser
+COPY signing-service                    /srv/http/
```
cameel: Please add a trailing `/` so that it's apparent that `signing_service/` is a directory.

containers/signing-service/Dockerfile.j2
```diff
@@ -0,0 +1,15 @@
...
+
+COPY testing.pref                       /etc/apt/preferences.d/
+COPY install-dependencies.sh            /tmp/
+RUN  /tmp/install-dependencies.sh
+
+COPY signing-service/requirements.lock  /srv/http/signing_service/
+COPY install-signing-service.sh         /tmp/
+RUN  /tmp/install-signing-service.sh
+
+USER signingserviceuser
```
cameel: Please call the user just `signing-service`.

containers/signing-service/install-dependencies.sh.j2
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash -e
+
+# Update package lists for Debian and install dependencies
+dependencies=(
+
+    # Packages needed to build python extensions
+    gcc
+    git
+    software-properties-common
```
cameel: Please reorganize the scripts a bit so that we can remove `gcc`, `git`, `software-properties-common` and `python-dev` after the build:
- Instead of running `install-dependencies.sh` directly from `Dockerfile`, call it from `install-signing-service.sh`.
- Create a separate list for dependencies that can be removed after the build
- Create a script that removes dependencies after the build and run it from `install-signing-service.sh`
    - Move `apt-get clean`and similar to that script

containers/signing-service/install-dependencies.sh.j2
```diff
@@ -0,0 +1,41 @@
...
+
+apt-get install              \
+    --target-release testing \
+    --assume-yes             \
+    --no-install-recommends  \
+    ${testing_dependencies[*]}
+
+# Add non-privileges user
+groupadd --gid 999 signingserviceuser
+useradd --system --uid 999 --gid signingserviceuser signingserviceuser
```
cameel: Looks good. Let's use it more.

But I'd put it in a separate script called `create-user.sh` and run it before everything else.

containers/signing-service/Dockerfile.j2
cameel: This is not a server app. Let's put it somewhere else. Let's use `/usr/lib/signing-service/`. It should be owned by root and read-only. Our user does not need to modify these files anyway.

cameel:
Also:
- This is too large for a single commit. Please split it into smaller ones:
    - Signing Service container
    - Signing Service package
    - README
- I have rewritten the README changes: https://github.com/golemfactory/concent-deployment/tree/readme-signing-service. Please cherry-pick it into your branch and then remove my branch.

README.md
```diff
@@ -241,3 +241,77 @@ docker run                         \
...
+   --rm                                                   \
+   --hostname signing-service                             \
+   --network  host                                        \
+   --name     signing-service                             \
+   --volume   <path to ethereum key>:/srv/http/.ethereum/ \
+   signing-service                                        \
+   <concent_cluster_address>
+```
+
+### Building the concent signing service with manually create docker for it.
```
michalpiasecki1: same as above

README.md
```diff
@@ -241,3 +241,77 @@ docker run                         \
...
+  This is necessary to be able to run docker commands without `root` privileges.
+  Note that the change will not take effect until you log out of the current shell session.
+
+```bash
+sudo groupadd docker
+sudo usermod --all --groups docker <user>
+```
+
+
+### Building the concent signing service with automatically create docker for it.
```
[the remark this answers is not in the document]
cameel: Yeah. I have rewritten the instructions (f61a5f9e1c62957045857d8dbc9ae78d423b255c), but @bartoszbetka has not cherry-picked it into this pull request yet.

README.md
cameel: Same as above. See f61a5f9e1c62957045857d8dbc9ae78d423b255c.

containers/Makefile
```diff
@@ -79,7 +89,22 @@ verifier: build/ $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/gol
 	docker build --tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) build/verifier/
 	docker tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) $(IMAGE_PREFIX)verifier:latest
 
-docker-clean: nginx-proxy-clean concent-api-clean postgresql-clean verifier-clean
+signing-service: build/ $(CONCENT_API_SOURCE) $(SIGNING_SERVICE_SOURCE) build/signing-service/signing-service/
```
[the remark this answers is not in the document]
bartoszbetka: The signing service uses `testing.pref`, which is located in` SIGNING_SERVICE_SOURCE_FROM_CONCENT_API`, but it is copied from `build/concent-api/testing.pref`. I change the solution to copy this file from `../concent-builder/files/testing.pref`, just like the concent-api application does.

containers/Makefile
```diff
@@ -79,7 +89,22 @@ verifier: build/ $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/gol
...
+	docker build --tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) build/signing-service/
+	docker tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) $(IMAGE_PREFIX)signing-service:latest
+
+concent-signing-service-package: build/ $(CONCENT_API_SOURCE) $(SIGNING_SERVICE_SOURCE) build/concent-signing-service-%.tar.xz
+
+build/concent-signing-service-%.tar.xz:  build/signing-service/signing-service/
+	tar                                                                                                         \
+		--create                                                                                                \
+		--verbose                                                                                               \
+		--xz                                                                                                    \
```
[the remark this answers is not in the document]
bartoszbetka: The indents are fine, github transforms it into a strange look.

containers/Makefile

bartoszbetka: Changed solution above, not require this target any more.

containers/Makefile

containers/Makefile
bartoszbetka: Outdated after changing the solution.

containers/Makefile
bartoszbetka: Outdated after changing the solution.

bartoszbetka: Outdated after changing the solution.

bartoszbetka: Outdated after changing the solution.

containers/signing-service/Dockerfile.j2
bartoszbetka: Done.

containers/signing-service/install-dependencies.sh.j2
bartoszbetka: Done. The solution must be checked before merged when sign-service will complete on concent side.

containers/signing-service/install-dependencies.sh.j2
bartoszbetka: Done.

containers/signing-service/Dockerfile.j2
bartoszbetka: Done.

containers/Makefile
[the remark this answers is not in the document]
cameel: OK

containers/Makefile
```diff
@@ -174,15 +211,17 @@ build/config-with-versions.yml: versions.yml config.yml ../tools/merge-yaml.py
...
-
-.PHONY:                                 \
-    all                                 \
-    nginx-proxy   nginx-proxy-clean     \
-    nginx-storage nginx-storage-clean   \
-    postgresql    postgresql-clean      \
-    concent-api   concent-api-clean     \
-    verifier      verifier-clean        \
-    docker-clean                        \
+	rm -rf build/ dist/
```
cameel: `clean` target is usually meant to remove only intermediate files.

There should be a separate `distclean` for removing `dist/`. See [GNU Coding Standards > Standard Targets for Users](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html).

containers/signing-service/install-dependencies.sh.j2
cameel: Please create a list of stuff that needs to be fixed before merging and put it in the pull request description. Otherwise it'll be easy to forget this stuff.

containers/signing-service/Dockerfile.j2
```diff
@@ -0,0 +1,19 @@
...
+COPY signing-service/requirements.lock  /usr/lib/signing-service/
+COPY testing.pref                       /etc/apt/preferences.d/
+COPY install-dependencies.sh            /tmp/
+COPY install-signing-service.sh         /tmp/
+COPY clean-container.sh                 /tmp/
+COPY middleman_protocol/                /usr/lib/middleman_protocol/
+RUN  /tmp/install-signing-service.sh
+
+USER signing-service
+COPY signing-service/                   /usr/lib/signing-service/
```
cameel: Please put both `middleman_protocol/` and `signing_service/` as subdirectories in a common directory. For example `/usr/lib/concent/` or `/usr/lib/signing_service/`. Put the virtualenv in that directory too.

We want to have only a single directory in `/usr/lib/` and everything else inside it.

containers/signing-service/clean-container.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -e
+
+temporary_dependencies=(
+
+    gcc
+    libssl-dev
+    git
+    pkg-config
```
cameel: We should deduplicate it. It's bad to have lists like this in two places. Maybe create two text files (one for temporary dependencies and one for permanent ones) and put dependencies in them? Or a common script sourced by both?

containers/signing-service/clean-container.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Clean up
+apt-get clean
+apt autoremove --assume-yes
+apt-get remove --purge --assume-yes ${temporary_dependencies[*]}
+
+rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Temporary file which is needed to build middleman protocol
+rm /usr/lib/RELEASE-VERSION
```
cameel: This file can stay.  But it should be in a common subdirectory with everything else we install, not directly in `/usr/lib/`.

containers/signing-service/install-signing-service.sh.j2
```diff
@@ -0,0 +1,17 @@
...
+virtualenv_dir=/usr/lib/virtualenv
+# Create virtualenv and Django app's dependencies
+virtualenv --python "python{{ python_version }}" "$virtualenv_dir"
+source "$virtualenv_dir/bin/activate"
+python -m pip install --upgrade pip
+python -m pip install --requirement "/usr/lib/signing-service/requirements.lock"
+(
+    cd /usr/lib/middleman_protocol/
+    python setup.py develop
+)
```
cameel: As discussed before, Signing Service should have the relative path to `middleman_protocol` hard-coded in its dependencies. Then you will not have to install it explicitly.

containers/signing-service/install-signing-service.sh.j2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+# Install signing-service dependencies
+/tmp/install-dependencies.sh
```
cameel: Please don't hard-code path to `/tmp/`. Use paths relative to script's location.

containers/Makefile
```diff
@@ -152,6 +168,15 @@ build/repositories/concent/: build/package-builder/clone-or-update.sh /var/conce
 build/repositories/golem/: build/package-builder/clone-or-update.sh /var/concent-data/golem
 	build/package-builder/clone-or-update.sh  "golem" "$@"
 
+build/signing-service/signing-service/RELEASE-VERSION: build/signing-service/signing-service/
+	echo -n "$(shell cd build/repositories/concent/; git describe --tags)" > "$@"
```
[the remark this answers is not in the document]
bartoszbetka: I generate this the same way as for Concent.

containers/Makefile

containers/signing-service/Dockerfile.j2
bartoszbetka: Done.

containers/signing-service/clean-container.sh
bartoszbetka: Done.

containers/signing-service/install-signing-service.sh.j2
bartoszbetka: Done.

containers/signing-service/clean-container.sh
bartoszbetka: Done.

containers/Makefile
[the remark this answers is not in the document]
cameel: Does not seems so:

https://github.com/golemfactory/concent-deployment/blob/549f43e1d2838068b3b49ab36569036646848969/containers/Makefile#L50

Versions for Concent always contain 16 characters from the hash if there's no tag:

containers/Makefile
```diff
@@ -152,6 +168,15 @@ build/repositories/concent/: build/package-builder/clone-or-update.sh /var/conce
...
+
+build/signing-service/signing-service/: build/repositories/concent/
+	cp --recursive "$<signing_service/." "$@"
```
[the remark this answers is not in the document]
cameel: If it will affect how the Makefile works, you can leave it as is. But if not, I'd prefer to change it for the sake of readability. I was very confused when I have seen this for the first time and it took me a while to realize that it's the `$<` concatenated with a string and not some weird make syntax.

What about something like this:
``` make
cp --recursive "$(abspath $</signing_service/)" "$@"
```

This may add an extra `/` but `$(abspath)` will remove it.

Anyway, if it's a problem then forget about it. Either way it looks weird at first.

containers/signing-service/install-dependencies.sh.j2
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
+
+# Update package lists for Debian and install dependencies
+# Packages needed to build python extensions, dependencies of golem-messages in particular
+readarray dependencies < ${BASH_SOURCE%/*}/packages.txt
+
+echo ${dependencies[@]}
```
cameel: Maybe also add a message saying that you're printing the dependencies?

Also, won't `apt-get` print them anyway during installation?

containers/signing-service/packages.txt
```diff
@@ -0,0 +1,7 @@
+gcc
```
cameel: I'd call the file `build-dependencies.txt` or something like this to make it clear that the file contains packages that will be uninstalled after the build.

We don't want someone to add his package here and discover later that the container does not work becuause an important dependency has been removed after the build.

containers/signing-service/Dockerfile.j2
```diff
@@ -0,0 +1,21 @@
+FROM debian:{{ debian_version }}
+ARG SIGNING_SERVICE_PATH=/usr/lib/signing_service
```
cameel: `ARG` suggests that the user can set it from the command line when running `docker build` but that's not the case here. We have the path hard-coded in scripts so the container will only work with this one value.

If you want to have variable here, I'd use `ENV` instead. But I think it's better to hard-code it if we don't support changing it.

containers/signing-service/clean-container.sh
cameel: The solution with `readarray` is great.

:+1:

containers/Makefile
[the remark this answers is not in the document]

containers/signing-service/install-signing-service.sh.j2
bartoszbetka: This solution must wait for [#736](https://github.com/golemfactory/concent/issues/736)

containers/signing-service/install-dependencies.sh.j2
bartoszbetka: Done.

containers/signing-service/packages.txt
bartoszbetka: Done.

containers/signing-service/Dockerfile.j2
bartoszbetka: Done.

containers/Makefile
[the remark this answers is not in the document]

containers/Makefile
```diff
@@ -152,6 +168,15 @@ build/repositories/concent/: build/package-builder/clone-or-update.sh /var/conce
...
+
+build/signing-service/middleman_protocol/: build/repositories/concent/
+	cp --recursive "$<middleman_protocol/." "$@"
```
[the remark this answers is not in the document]

README.md
```diff
@@ -241,3 +241,82 @@ docker run                         \
...
 ```
 
+## Concent Signing Service
+In addition to Concent itself, this repository contains files necessary to build the Concent Signing Service.
+`Makefile` builds a Docker container but also produces a source package that includes `Dockerfile` and all source files needed to build it.
+The package can be used to build the container without having to set up `concent-deployment`.
+
+### Building the Signing Service package
+This is only needed if you want to build the package yourself.
+If you have received the package and want to build and run it, skip this section.
```
cameel: Please add a little tweak before merging:

``` markdown
This is only needed if you want to build the package yourself.
If you have received the package and only want to build and run the docker container, you can skip this section.
```

README.md
```diff
@@ -300,7 +300,7 @@ docker run
     --network     host                                                           \
     --hostname    signing-service                                                \
     --name        signing-service                                                \
-    --volume      contract.key:/etc/contract.key:ro                              \
+    --volume      < path to file >/contract.key:/etc/contract.key:ro             \
```
cameel: I don't think this change is necessary. `contract.key` is a relative path and it's explained below that it's just one of the possible locations:

> - There's a file called `contract.key` in your local directory and that it contains the private key that can sign Ethereum transactions from the account owned by Concent contract.

README.md
```diff
@@ -295,18 +295,19 @@ If you have received the package and only want to build and run the docker conta
...
+docker run                                                                               \
+    --rm                                                                                 \
+    --network     host                                                                   \
+    --hostname    signing-service                                                        \
+    --name        signing-service                                                        \
+    --restart     on-failure                                                             \
+    concent-signing-service                                                              \
+        concent.golem.network                              #host                         \
+        85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us   #concent_public_key in base64 \
+        --concent-cluster-port                  9055                                     \             \
```
cameel: You have two `\` characters here.

README.md
```diff
@@ -295,18 +295,19 @@ If you have received the package and only want to build and run the docker conta
...
-        --dsn                       https://1234:5678@talkback.golem.network/555 \
-        --ethereum-private-key-path /etc/contract.key
+docker run                                                                               \
+    --rm                                                                                 \
+    --network     host                                                                   \
+    --hostname    signing-service                                                        \
+    --name        signing-service                                                        \
+    --restart     on-failure                                                             \
+    concent-signing-service                                                              \
+        concent.golem.network                              #host                         \
```
cameel: I think that
```
--host                      concent.golem.network
```

was better than having a comment like that.

README.md
```diff
@@ -295,18 +295,19 @@ If you have received the package and only want to build and run the docker conta
...
+        85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us   #concent_public_key in base64 \
```
cameel: There's a section below that explains the parameters. You should add your remarks there instead of adding comments here.

E.g.:
``` markdown
This asumes that:
(...)
- `85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us` is Concent's public key encoded in base64.
(...)
```

README.md
```diff
@@ -295,18 +295,19 @@ If you have received the package and only want to build and run the docker conta
...
+        --concent-cluster-port                  9055                                     \             \
+        --ethereum-private-key-from-env                                                  \
+        --signing-service-private-key-from-env                                           \
+        --sentry-dsn-from-env
```
cameel: I you're showing the `...-from-env` options you should also show how to set the corresponding variables in a docker call.

README.md
bartoszbetka: Done

README.md
bartoszbetka: The `--host` flag not exist.

README.md

README.md
```diff
@@ -292,31 +292,33 @@ If you have received the package and only want to build and run the docker conta
     ```
 
 ### Running the Signing Service
+Set `ETHEREUM_RPIVATE_KEY`, `SIGNING_SERVICE_PRIVATE_KEY` and if you want `SENTRY_DSN` to bash variables encoded in base64 before proceeding.
```
cameel: Please reword and put it in the section after the command:
```
This assumes that:
(...)
- There's a shell variable called `ETHEREUM_RPIVATE_KEY` and it contains base64-encoded private key of the Concent contract.
- There's a shell variable called `SIGNING_SERVICE_PRIVATE_KEY` and it contains base64-encoded private key for signing Golem Messages created by the Signing Service.
- There's a shell variable called `SENTRY_DSN` that contains the secret ID that allows submitting crash report to a [Sentry](https://sentry.io) project.
(...)
```

README.md
```diff
@@ -292,31 +292,33 @@ If you have received the package and only want to build and run the docker conta
...
     --network     host                                                                   \
     --hostname    signing-service                                                        \
     --name        signing-service                                                        \
     --restart     on-failure                                                             \
     concent-signing-service                                                              \
-        concent.golem.network                              #host                         \
-        85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us   #concent_public_key in base64 \
-        --concent-cluster-port                  9055                                     \             \
+        concent.golem.network                                                            \
+        85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us                                 \
```
cameel: Please remove excessive whitespace now that it's no longer needed.

README.md
```diff
@@ -292,31 +292,33 @@ If you have received the package and only want to build and run the docker conta
...
         --ethereum-private-key-from-env                                                  \
         --signing-service-private-key-from-env                                           \
         --sentry-dsn-from-env
 ```
 
 This assumes that:
 - The service can connect to a Concent cluster at `concent.golem.network:9055`.
-- There's a file called `contract.key` in your local directory and that it contains the private key that can sign Ethereum transactions from the account owned by Concent contract.
+- `85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us` is Concent's public key encoded in base64.
+- The `--ethereum-private-key-from-env` and `--signing-service-private-key-from-env` are required.
```
cameel: > The `--ethereum-private-key-from-env` and `--signing-service-private-key-from-env` are required.

This not necessary. We're only showing an example here. The person setting it up can always see the full set of options using `--help`.

README.md
```diff
@@ -292,31 +292,33 @@ If you have received the package and only want to build and run the docker conta
...
 - The service can report errors to a [Sentry](https://sentry.io) instance.
-    The one given above is just an example.
-    If so, you need to provide a valid DSN for the project that should receive the reports.
-    Otherwise just skip the `--dsn` parameter.
+    This is not needed setting, you can skip `--sentry-dsn-from-env` and `env SENTRY_DSN` parameters.
```
cameel: Please restore the first two lines. Change only the third (``Otherwise just skip the `--dsn` parameter.``):

```
Otherwise just skip the `--sentry-dsn-from-env` parameter and the `SENTRY_DSN` variable.
```

README.md
bartoszbetka: Done

README.md

README.md

README.md
cameel: OK

README.md

README.md
```diff
@@ -241,3 +241,89 @@ docker run                         \
...
+This assumes that:
+- The service can connect to a Concent cluster at `concent.golem.network:9055`.
+- `85cZzVjahnRpUBwm0zlNnqTdYom1LF1P1WNShLg17cmhN2Us` is Concent's public key encoded in base64.
+- There's a shell variable called `ETHEREUM_RPIVATE_KEY` and it contains base64-encoded private key of the Concent contract.
+- There's a shell variable called `SIGNING_SERVICE_PRIVATE_KEY` and it contains base64-encoded private key for signing Golem Messages created by the Signing Service.
+- There's a shell variable called `SENTRY_DSN` that contains the secret ID that allows submitting crash report to a [Sentry](https://sentry.io) project.
+    The one given above is just an example.
+    If so, you need to provide a valid DSN for the project that should receive the reports.
+    Otherwise just skip the `--sentry-dsn-from-env` parameter and the `SENTRY_DSN` variable.
+- The `--sentry-environment` variable add tag to sentry which specify in our case cluster name.
```

```
- `mainnet` is the name of the Sentry environment that should be included in error reports.
```

Issue #193, opened 2018-07-20
Refactor container Makefile to reduce repetition


cameel:
`containers/Makefile` has several instances of code that almost identical and could be deduplicated using a pattern rule:

https://github.com/golemfactory/concent-deployment/blob/3bb35a3c74acf434a97e3b39320e26408c21f5d4/containers/Makefile#L62-L80

https://github.com/golemfactory/concent-deployment/blob/3bb35a3c74acf434a97e3b39320e26408c21f5d4/containers/Makefile#L84-L97

Please deduplicate it.

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/159193193

PR #199, opened 2018-07-27
build dir as order only prerequisite, docker rmi deduplicated, docker…

michalpiasecki1:
docker build and docker tag deduplicated
build dir as order only prerequisite, 
docker rmi deduplicated

Resolves: https://github.com/golemfactory/concent-deployment/issues/194, https://github.com/golemfactory/concent-deployment/issues/193

PR #211, opened 2018-08-14
Chore refactor makefile to reduce repetition

ghost:
new pull request due to wrong branch name, and comments  in https://github.com/golemfactory/concent-deployment/pull/199 ( particulary https://github.com/golemfactory/concent-deployment/pull/199; #pullrequestreview-144854386 ). Solves issues found in previous PR as: separate commits, separate PR, chore in branch name, phony target and static pattern rules

1. docker rmi deduplicated
2. static pattern rules introduced
3.  %-rmi is a phony target

Resolves: #193

### DIFF
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -1,49 +1,49 @@
-NGINX_PROXY_SOURCE :=               \
-  build/nginx-proxy/Dockerfile      \
-  build/nginx-proxy/index.html      \
-  build/nginx-proxy/error-pages.conf
+NGINX_PROXY_SOURCE :=                 \
+    build/nginx-proxy/Dockerfile      \
+    build/nginx-proxy/index.html      \
+    build/nginx-proxy/error-pages.conf
 
 NGINX_STORAGE_SOURCE_FROM_NGINX_PROXY := \
-  build/nginx-storage/error-pages.conf
+    build/nginx-storage/error-pages.conf
 
-NGINX_STORAGE_SOURCE :=                            \
-  build/nginx-storage/Dockerfile                   \
-  build/nginx-storage/install-dependencies.sh      \
-  build/nginx-storage/table_utils.lua              \
-  build/nginx-storage/file-transfer-config.lua     \
-  build/nginx-storage/file-transfer-endpoints.lua  \
-  build/nginx-storage/file-transfer-endpoints.conf \
-  build/nginx-storage/default.conf                 \
+NGINX_STORAGE_SOURCE :=                              \
+    build/nginx-storage/Dockerfile                   \
+    build/nginx-storage/install-dependencies.sh      \
+    build/nginx-storage/table_utils.lua              \
+    build/nginx-storage/file-transfer-config.lua     \
+    build/nginx-storage/file-transfer-endpoints.lua  \
+    build/nginx-storage/file-transfer-endpoints.conf \
+    build/nginx-storage/default.conf                 \
  $(NGINX_STORAGE_SOURCE_FROM_NGINX_PROXY)
 
-POSTGRESQL_SOURCE :=                  \
-  build/postgresql/Dockerfile         \
-  build/postgresql/delete-database.sh \
-  build/postgresql/create-database.sh
+POSTGRESQL_SOURCE :=                    \
+    build/postgresql/Dockerfile         \
+    build/postgresql/delete-database.sh \
+    build/postgresql/create-database.sh
 
-CONCENT_API_SOURCE :=                       \
-  build/concent-api/install-concent-api.sh  \
-  build/concent-api/install-dependencies.sh \
-  build/concent-api/Dockerfile              \
-  build/concent-api/concent-api-manage.sh   \
-  build/concent-api/run-celery.sh           \
-  build/concent-api/run-gunicorn.sh         \
-  build/concent-api/local_settings.py
+CONCENT_API_SOURCE :=                         \
+    build/concent-api/install-concent-api.sh  \
+    build/concent-api/install-dependencies.sh \
+    build/concent-api/Dockerfile              \
+    build/concent-api/concent-api-manage.sh   \
+    build/concent-api/run-celery.sh           \
+    build/concent-api/run-gunicorn.sh         \
+    build/concent-api/local_settings.py
 
-VERIFIER_SOURCE_FROM_CONCENT_API :=     \
-  build/verifier/install-concent-api.sh \
-  build/verifier/concent-api-manage.sh  \
-  build/verifier/run-celery.sh          \
-  build/verifier/testing.pref           \
-  build/verifier/requirements.lock      \
-  build/verifier/concent-api.tar        \
-  build/verifier/local_settings.py
+VERIFIER_SOURCE_FROM_CONCENT_API :=       \
+    build/verifier/install-concent-api.sh \
+    build/verifier/concent-api-manage.sh  \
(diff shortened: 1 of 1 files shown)


containers/Makefile
```diff
@@ -56,28 +56,42 @@ ifndef IMAGE_PREFIX
...
+    build/concent-api/testing.pref      							\
+    build/repositories/concent/
+
+postgresql!dependencies: 															\
+    $(POSTGRESQL_SOURCE)
+
+verifier!dependencies:																\
+		$(VERIFIER_SOURCE) 																\
+		build/repositories/golem/ 												\
+		build/verifier/golem/scripts/ 										\
```
cameel: Please fix the whitespace here.

- You're inserting tabs everywhere. Even inside the text (e.g. `build/verifier/golem` and `\`). We're using spaces everywhere, please use them too. Otherwise the code looks very misaligned.
    - The only place where we use tabs is at the beginning of a line containing a bash command in a target. And that's only because make syntax requires it.
- It's fine to add a little whitespace before `\` but this is way too much. There's so much that I have to scroll the diff horizontally to see that there's `\` at the end of some lines.
- `\` should all be either in the same column (preferred) or immediately after the text.

I have just realized that we don't have this written down in our style yet so I have added a new proposal. Please see https://github.com/Code-Poets/coding-style/issues/16

containers/Makefile
```diff
@@ -56,28 +56,42 @@ ifndef IMAGE_PREFIX
...
+		build/verifier/golem/imgverifier-requirements.txt
```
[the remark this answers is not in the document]
michalpiasecki1: most of the file lines uses 4 spaces as indent (55 occurences), some one tab (55 occurences) some two tabs (29 occurences) and some two spaces (3 occurences). I used two tabs as in mine editor tab=two spaces. To make it consistent, I propose to change every indent to one tab.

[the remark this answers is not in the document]
cameel: Have you seen https://github.com/golemfactory/concent-deployment/pull/211#discussion_r209999469?

containers/Makefile
michalpiasecki1: yeah, the problem is with different rendering engine in github and atom, apparently tabs in atom does not have fixed length, and slashes appeared to be alligned, definetely within screen width

michalpiasecki1: excessive tabs replaced with spaces

containers/Makefile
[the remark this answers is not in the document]
michalpiasecki1: yes, replied and fixed

containers/Makefile
```diff
@@ -114,22 +114,21 @@ build/concent-api/concent-api.tar: build/repositories/concent/  build/virtualenv
...
-	build/package-builder/build-concent-api-docs.sh   \
-		"build/repositories/concent/"                 \
-		"$(abspath package-builder)"                  \
-		"$(abspath build)"                            \
-		"$(dir $@)"
+	build/package-builder/build-concent-api-docs.sh\
+	  "build/repositories/concent/"                \
+	  "$(abspath package-builder)"                 \
+	  "$(abspath build)"                           \
+	  "$(dir $@)"
```
cameel: You have replaced all indents with 2-space ones. As you can see from the diff, we're already using 4-space indents. Changing it in the middle of an unrelated pull request is a bad idea.

Please keep the original indentation. Just fix the code you have added.

I've updated the issue about coding style to add info about 4-character spaces.

containers/Makefile
```diff
@@ -114,22 +114,21 @@ build/concent-api/concent-api.tar: build/repositories/concent/  build/virtualenv
...
 
 build/nginx-proxy/concent-api-docs.tar: build/repositories/concent/ build/package-builder/build-concent-api-docs.sh
 	mkdir --parents $(dir $@)
-	build/package-builder/build-concent-api-docs.sh   \
-		"build/repositories/concent/"                 \
-		"$(abspath package-builder)"                  \
-		"$(abspath build)"                            \
-		"$(dir $@)"
+	build/package-builder/build-concent-api-docs.sh\
+	  "build/repositories/concent/"                \
```
cameel: It's a good practice to always leave a space before `\`. Otherwise it's easy to run into problems like this:
``` bash
cat\
a.txt\
b.txt\
c.txt
```
```
bash: cata.txtb.txtc.txt: command not found
```

That's because bash does not add spaces on its own and sees your command as `cata.txtb.txtc.txt` rather than `cat a.txt b.txt c.txt`.

There is some whitespace at the beginning of the line in this case which would prevent this problem but that's not always the case so, as a rule, we always add at least one space at the end.

I've updated the issue for updating coding style to make it clear.

containers/Makefile
```diff
@@ -164,28 +160,27 @@ build/%: %
...
+.PHONY:                            \
+	all                              \
+	nginx-proxy   nginx-proxy-clean  \
+	nginx-storage nginx-storage-clean\
+	postgresql    postgresql-clean   \
+	concent-api   concent-api-clean  \
+	verifier      verifier-clean     \
+	docker-clean                     \
+	clean                            \
+	%-rmi
```
cameel: Are you sure this works? I think that pattern rules will only work if `%` is in the target as well. But I may be wrong. Please check.

containers/Makefile
```diff
@@ -164,28 +160,27 @@ build/%: %
...
 build/config-with-versions.yml: versions.yml config.yml ../tools/merge-yaml.py
-	../tools/merge-yaml.py                     \
-		"$<"                                   \
-		"config.yml"                           \
+	../tools/merge-yaml.py\
+		"$<"                \
+		"config.yml"        \
 		--output "$@"
 
-
```
cameel: Please avoid unnecessary changes, like removing these empty lines.

I'm not saying that removing them is a bad idea but I don't want you and @bartoszbetka to get into a loop where you remove them and he adds them back :)

When editing an existing file please adapt to the style already used in it unless the issue is specifically about refactoring the style.

containers/Makefile
```diff
@@ -117,18 +117,18 @@ build/concent-api/concent-api.tar: build/repositories/concent/  build/virtualenv
 build/nginx-proxy/concent-api-assets.tar: build/repositories/concent/ build/virtualenv/ build/package-builder/build-concent-api-assets.sh
 	mkdir --parents "$(dir $@)"
 	build/package-builder/build-concent-api-assets.sh\
-		"build/repositories/concent/"                  \
-		"$(abspath package-builder)"                   \
-		"$(abspath build)"                             \
-		"$(dir $@)"
+	  "build/repositories/concent/"                  \
```
cameel: I'm not sure if having a mix of tabs and spaces at the beginning of a line is a good idea.

containers/Makefile
michalpiasecki1: It was the only one occurrence of two empty lines in a row, so it looked more than correcting a mistake than imposing own coding style. Fully agree about adapting to existing style.

containers/Makefile
michalpiasecki1: That is correct,  I did some tests and .PHONY targets must be declared explicitly.

containers/Makefile
michalpiasecki1: according to coding style, I used tab where is needed by makefile syntax, and space everywhere else (shell arguments)

cameel:
Could you look at the diff and ensure that there are no unnecessary trivial whitespace changes?

PR #223, opened 2018-08-29
Feature change paths for building and deploying system to home user path on the concent builder server

bartoszbetka:


concent-builder/install-secrets.yml
```diff
@@ -0,0 +1,50 @@
+- hosts:
+    - concent-builder
+  vars_files:
```
cameel: Moving secrets from `install-repositories.yml` to `install-secrets.yml` should be in a separate commit.

Also, you should update README so that it says to run this playbook.

concent-builder/build-test-and-push.yml
```diff
@@ -29,27 +27,23 @@
           copy:
             src:   "{{ deployment_values }}/{{ item }}"
             dest:  "{{ build_dir }}/concent-deployment/kubernetes/"
-            owner: "{{ shared_user }}"
-            group: "{{ shared_user }}"
```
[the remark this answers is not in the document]
bartoszbetka: This block has a `always` directive.

concent-builder/build-test-and-push.yml
```diff
@@ -6,9 +6,7 @@
     - "{{ deployment_values }}/var.yml"
     - "{{ deployment_values }}/var-{{ cluster }}.yml"
   tasks:
-    - become:      yes
-      become_user: "{{ shared_user }}"
-      block:
+    - block:
```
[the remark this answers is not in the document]
bartoszbetka: Done.

concent-builder/install-secrets.yml
bartoszbetka: Done.

README.md
```diff
@@ -185,8 +192,7 @@ ansible-playbook migrate-db.yml                                    \
 cd concent-deployment/concent-builder/
 ansible-playbook deploy.yml                                        \
     --extra-vars cluster=$cluster                                  \
-    --inventory  ../../concent-deployment-values/ansible_inventory \
-    --user       $user
+    --inventory  ../../inventory
```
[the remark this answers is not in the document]
cameel: @bartoszbetka What about this change?

concent-builder/build-test-and-push.yml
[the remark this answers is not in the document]
cameel: Where? I don't see it in the file.

concent-builder/install-repositories.yml
```diff
@@ -20,45 +20,3 @@
             # FIXME: Do not blindly accept the hostkey
             accept_hostkey: yes
           with_dict:        "{{ repositories }}"
-
```
cameel: This file no longer needs to include `var-{{ cluster }}.yml"`

README.md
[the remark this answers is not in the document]
cameel: After discussing this, the issue was that with `ansible_inventory` the playbook would not work when run from inside the `concent-builder` itself. It would try to use the external IP and fail. The solution is to add `--connection=local` when running from `concent-builder`.

But the fact that it fails when using the external IP is not the typical situation. It's a special case that happens only if you're trying to configure a server while running from that server. Let's keep the instructions generic, with `ansible_inventory`. We could at most add a note somewhere that `--connection=local` exists. For example we could replace this in one of the previous sections:

``` markdown
All the instructions below assume that you're using the remote server.
```

with

``` markdown
All the instructions below assume that you're using local playbooks to run build and deployment commands on the remote server.

Note that if you're running the playbooks themselves from within that server too, you need to add `--connection=local` to your `ansible-playbook` calls.
Otherwise Ansible will run its commands over SSH (rather than directly) using the public IP specified in `ansible_inventory`, which will likely fail because you're not supposed to have your private SSH key on the remote server you're connecting to.
```

concent-builder/install-secrets.yml
```diff
@@ -0,0 +1,50 @@
...
+    - "{{ deployment_values }}/var.yml"
+    - "{{ deployment_values }}/var-{{ cluster }}.yml"
+  tasks:
+    - become:      yes
+      become_user: "{{ shared_user }}"
+      block:
+        - name: Create a directory for secrets
+          file:
+            path:  "{{ data_dir }}/concent-secrets/{{ cluster }}/"
+            state: directory
```
cameel: Can this directory really be created by `shared_user`? Doesn't root have to do this?

bartoszbetka: No, because we use builder user in cloud management playbooks. The question is, do we want to change it?

cameel: Actually I've seen this later in another of your pull requests and I realized that you rely on the fact that `{{ data_dir }}` already exists  and is owned by `shared_user` - `configure.yml` creates it. So it's fine as is. You could just add a comment that says this.

bartoszbetka: The problem is with cloud management playbooks. They use `builder` user, we must adapt it to the current solution. 
Two solution that I see:
- Change they, to building and deploying in home user directory
- Add installation of concent-deployment repository to builder user and then build and deploy as builder user.

cameel: Which one do you think will require less changes? I'm still undecided but if one of much easier than the other then we could do it for now.

bartoszbetka: I think that the first one requires less changes.

cameel: OK. Let's change it to build in home directory. In fact. let's just remove `builder` if it's not used for anything else.

PR #226, opened 2018-09-03
Add separate cloud management from cluster deployment

### LINKED ISSUE #198: Separate cloud management from cluster deployment
#### Overview
Concent stores its data on a [PostgreSQL database server managed by Cloud SQL](https://cloud.google.com/sql/docs/postgres/). Creation and removal of a database is a part of a cluster deployment process.

PostgreSQL has an extensive [privilege system](https://www.postgresql.org/docs/current/static/sql-grant.html) but there's no way to grant permissions to a specific database while it does not exist. This means that to create a database during deployment one needs to be able to create or delete arbitrary databases, also those belonging to other clusters. Currently we grant this privilege only to the database administrator (`postgres` role) which means that even someone deploying to the `dev` cluster needs to know superuser credentials.

The goal of this task is to separate database management from cluster deployment. All databases should be created ahead of time by someone with access to all the clusters. Cluster deployment process should be altered so that it clears an existing database and reuses it instead of dropping it and recreating it. This will make it possible to allow someone to deploy a single cluster without letting him access databases belonging to other clusters.

Managing databases is only one example of activities that I refer to as "cloud management". These include also creating/deleting persistent disks, creating/deleting clusters, reserving IPs. Basically, anything that requires access to the whole Google Cloud project rather than just to a specific Kubernetes cluster within that project. All of those things will later be scripted but right now we do them manually - that's why in this issue we focus solely on the database management.

#### Secrets
We want to separate cluster-level secrets from cloud-level secrets. Currently the cluster deployment scripts assume that there's one secret directory per cluster and it contains a file called `var-secret.yml` with the following fields:
- `db_admin_password`
- `db_user_password`
- `django_admin_password_hash`
- `rabbitmq_password`

From now on we want to have a separate directory for cloud-level secrets. This directory should have a file called `cloud-secrets.yml` and `db_admin_password` should be moved to that file. `var-secret.yml` should be renamed to `cluster-secrets.yml` and contain all the other fields.

Eventually build server will contain only `cluster-secrets.yml` for every test cluster (`dev`, `staging`, `testnet`). `cloud-secrets.yml` together with `cluster-secrets.yml` for `mainnet` cluster will only be available on a special machine used only for deployment.

**Do not** upload `cloud-secrets.yml` to the build server in `install-repositories.yml` playbook. Only the `cluster-secrets.yml` will now be needed for cluster deployment.

#### Cloud management playbooks
1. Create a directory called `cloud/` in the root directory of the repository. The directory will contain all the cloud management scripts and playbooks.
2. Create the following playbooks inside the new directory:
    - `create-databases.yml`: creates all databases and roles listed in the cluster configuration (`var-xxx.yml`).
    - `drop-databases.yml`: drops all databases and roles listed in the cluster configuration (`var-xxx.yml`).
    - If possible, reuse code from existing playbooks for `concent-builder`.

These playbooks will eventually be executed against a special deployment machine but for now assume that they're running on the build server. You may need to add the build server to the Cloud SQL IP whitelist  for now.

#### Modify cluster deployment
1. Replace the [create-database.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/containers/postgresql/create-database.sh) and [delete-database.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/containers/postgresql/delete-database.sh) scripts in the `postgresql` container with one called `reset-database.sh`.
    - The script should clear an existing database by dropping all [schema](https://www.postgresql.org/docs/10/static/ddl-schemas.html)s and creating an empty schema called `public` (the default schema in PostgreSQL).
2. Replace the [create-database](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/kubernetes/jobs/create-database.yml.j2) and [delete-database](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/kubernetes/jobs/delete-database.yml.j2) jobs with one called `reset-database`. It should use `reset-database.sh` to clear an existing database.
3. Remove `db_admin_password` from [db-secrets](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/kubernetes/secrets/db-secrets.yml.j2) secret and modify the jobs to use `db_user_password` (and the user role) instead.
4. Merge [create-job-dependencies.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/kubernetes/create-job-dependencies.sh) and [delete-job-dependencies.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/kubernetes/delete-job-dependencies.sh) with [create-config-maps.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/kubernetes/create-config-maps.sh.j2) and [delete-config-maps.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/kubernetes/delete-config-maps.sh). We no longer need to be able to create and delete them separately. The secrets used by jobs no longer contain the admin key for the database and can stay on the cluster.

#### Modify deployment playbooks
In this issue we assume that cluster management is still done from the build server. Eventually we want to have a separate machine just for storing the production secrets and deployment but this is out of scope of this particular issue.

1. Replace [create-db.yml](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/concent-builder/create-db.yml) and [delete-db.yml](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/concent-builder/delete-db.yml) with `reset-db.yml`.
2. Update deployment instructions in `README`.
    - Say that creating the databases is a separate step and requires access to the whole Google Cloud project. Normal cluster deployment only involves a reset or a migration.

#### `create-databases.yml` playbook
Parameters:
- `cluster`: name of the cluster, e.g. `concent-dev`.

The playbook should load the YAML files containing both cloud and cluster secrets.

Behavior:
- Use PostgreSQL client to create the database and role for the control cluster using values from `control_database_yyy` variables in `var-xxx.yml` and passwords from cluster and cloud secrets.
- Do the same for the storage database.
    - Create a helper playbook or a role that can be imported to avoid duplicated code.

##### Helper role
This should work in a similar way to how [create-database.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/containers/postgresql/create-database.sh) works now but the privileges should be a little different.

Parameters:
- Database host: IP of our Cloud SQL instance
- Database name
- Role name
- Role password
- Admin password

Passwords should never be passed directly to shell commands. Pass them via shell variables or files.

Behavior
1. Create database if it does not exist yet.
2. Create role if it does not exist yet.
    - Should own the new database and everything inside
    - Should **not** be able to create or drop databases (including its own database)
    - Should be able to create and destroy schemas inside its own database.

See [GRANT](https://www.postgresql.org/docs/current/static/sql-grant.html) for information about adding and revoking privileges.

#### `drop-databases.yml` playbook
Parameters:
- `cluster`: name of the cluster, e.g. `concent-dev`.

The playbook should load the YAML files containing both cloud and cluster secrets.

Behavior:
- Use PostgreSQL client to drop the database and role for the control cluster using values from `control_database_yyy`.
- Do the same for the storage database.
    - Create a helper playbook or a role that can be imported to avoid duplicated code.

##### Helper role
This should work in a similar way to how [delete-database.sh](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/containers/postgresql/delete-database.sh) works now.

Parameters:
- Database host: IP of our Cloud SQL instance
- Database name
- Role name
- Admin password

Behavior
1. Drop the database if it exists.
2. Delete the role if it exists.

bartoszbetka:
Resolves #198


README.md
```diff
@@ -157,10 +157,33 @@ ansible-playbook build-test-and-push.yml                           \
...
+    --user       $user
+```
+
+### Create or delete the database
+This is a separate step that require access to the whole Google Cloud project to modify CloudSQL database.
+In normal scenario, we don't needed to create or delete database. Normal cluster deployment only involves a reset or a migration.
+
+```bash
+cd concent-deployment/cloud/
+ansible-playbook create-databases.yml                              \
```
michalpiasecki1: if there will be only one database, there could be singular instead of plural in playbook names

cloud/drop-databases.yml
```diff
@@ -0,0 +1,13 @@
+- hosts:
+    - concent-builder
+  vars_files:
+    - ../concent-builder/consts.yml
+    - "{{ deployment_values }}/var.yml"
+    - "{{ deployment_values }}/var-{{ cluster }}.yml"
+    - "{{ local_secret_dir }}/{{ cluster }}/cluster-secrets.yml"
+    - "{{ local_secret_dir }}/cloud/cloud-secrets.yml"
+  vars:
+    choose_database_option: delete
```
michalpiasecki1: create and drop databases playbooks differ with one line only, is it good idea to create one file "manage-database" and pass parameter --extra-vars "database_option=delete"?

cloud/roles/postgres_connection/tasks/create-databases.yml
```diff
@@ -0,0 +1,42 @@
...
+  postgresql_privs:
+    login_host:      "{{ control_database_host }}"
+    login_password:  "{{ control_db_admin_password | b64decode }}"
+    db:              "{{ vars[ cluster_type + '_database_name'] }}"
+    role:            "{{ vars[ cluster_type + '_database_user'] }}"
+    type:            database
+    objs:            "{{ vars[ cluster_type + '_database_name'] }}"
+    privs:           ALL
+
+- name: Add postgres to membership of the "{{ vars[ cluster_type + '_database_user'] }}" role
```
michalpiasecki1: I don't get it. Can you rephrase? "add '_database_user' role postgres"?

cloud/roles/postgres_connection/tasks/delete-databases.yml
```diff
@@ -0,0 +1,28 @@
+- name: Revoke privileges for the "{{ vars[ cluster_type + '_database_user'] }}" role to the "{{ vars[ cluster_type + '_database_name'] }}" database
```
michalpiasecki1: Is this necessary ? If I am not wrong, it is like cleaning up before dropping an A-bomb

concent-builder/migrate-db.yml
```diff
@@ -14,16 +14,6 @@
...
-          args:
-            chdir: "{{ build_dir}}/concent-deployment/kubernetes/build/"
-
-        - name:    Create up to date secrets and config maps required by jobs
-          command: "{{ build_dir }}/concent-deployment/kubernetes/build/create-job-dependencies.sh {{ cluster_type }}"
-          args:
-            chdir: "{{ build_dir}}/concent-deployment/kubernetes/build/"
-
         - name:    Migrate the database used by concent-api
           command: kubectl create \
```
michalpiasecki1: maybe https://docs.ansible.com/ansible/2.5/modules/k8s_raw_module.html here?

README.md
cameel: There are two. `storage` and `control`.

cloud/drop-databases.yml
```diff
@@ -0,0 +1,13 @@
...
+  roles:
+    - { role: postgres_connection, cluster_type: control}
+    - { role: postgres_connection, cluster_type: storage}
```
cameel: Please fix style here.

cloud/drop-databases.yml
cameel: I'm a bit conflicted on this. Yeah, it's a bit of duplication but this playbook is short and all boilerplate anyway. It just starts a role which already deduplicates the parts that matter.

bartoszbetka: This code was duplicated intentionally in two playbooks to make deployment easier without extra option. I want to clarify two separate action.

cloud/roles/postgres_connection/tasks/create-databases.yml
bartoszbetka: Why don't you understand it? https://www.postgresql.org/docs/9.0/static/role-membership.html. Maybe change to member of role ? This phrase `"grant '_database_user' role postgres"` isn't understand for me.

cloud/roles/postgres_connection/tasks/delete-databases.yml
bartoszbetka: Done.

concent-builder/migrate-db.yml
bartoszbetka: This should be separate issue to research and change in all playbooks. @cameel What are you think about it?

cloud/drop-databases.yml
bartoszbetka: The style is the same as for python ?
```python
{
    'key1': 'value1',
    'key2': 'value2',
}
```

cloud/roles/postgres_connection/tasks/create-databases.yml
michalpiasecki1: add postgres to role '_database_user` ?

bartoszbetka: Done.

concent-builder/migrate-db.yml
cameel: I think we want to stick to our scripts because we'll be replacing it with Helm soon anyway.

cloud/create-databases.yml
```diff
@@ -0,0 +1,19 @@
+- hosts:
+    - concent-builder
+  vars_files:
+    - ../concent-builder/consts.yml
+    - "{{ deployment_values }}/var.yml"
+    - "{{ deployment_values }}/var-{{ cluster }}.yml"
+    - "{{ local_secret_dir }}/{{ cluster }}/cluster-secrets.yml"
+    - "{{ local_secret_dir }}/cloud/cloud-secrets.yml"
+  vars:
+    choose_database_option: create
```
cameel: Please rename this to `database_operation`

cloud/create-databases.yml
```diff
@@ -0,0 +1,19 @@
...
+  roles:
+    - {
+        role: postgres_connection,
+        cluster_type: control,
```
cameel: I'd make it more compact:

``` yaml
    - {role: postgres_connection, cluster_type: control}
    - {role: postgres_connection, cluster_type: storage}
```

cloud/create-databases.yml
```diff
@@ -0,0 +1,19 @@
+- hosts:
+    - concent-builder
+  vars_files:
+    - ../concent-builder/consts.yml
+    - "{{ deployment_values }}/var.yml"
+    - "{{ deployment_values }}/var-{{ cluster }}.yml"
+    - "{{ local_secret_dir }}/{{ cluster }}/cluster-secrets.yml"
```
cameel: I'm not sure if it's possible in Ansible but it would be great if we could have a vars file that accepts `cluster` as a parameter and includes all vars, constants and secrets for that cluster. This would reduce this to a single line.

@michalpiasecki1 Do you know any mechanism that would allow us to do this in a clean way?

containers/postgresql/reset-database.sh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+database_query="SELECT DISTINCT schema_name
+                FROM information_schema.schemata
+                WHERE schema_owner = '"$db_user"';"
+
+schema_names="$(
+        psql                                 \
+            "$db_name"                       \
+            --host "$db_host"                \
```
cameel: Please use 4-space indents.

containers/postgresql/reset-database.sh
```diff
@@ -0,0 +1,45 @@
...
+            --username $db_user              \
+            --tuples-only                    \
+            --command "$database_query"      \
+)"
+result=$?
+
+if [[ $result != 0 ]]; then
+    # Check if query has failed
```
cameel: This comment would not be necessary if your variable had a better name. E.g. `query_success` instead of just `result`.

concent-builder/configure.yml
```diff
@@ -73,6 +73,8 @@
             - libpq-dev
             - gcc
             - python3-yaml
+            - postgresql-9.6
+            - python-psycopg2
```
cameel: I think you should read the PostgreSQL version from `versions.yml`.

cloud/create-databases.yml
bartoszbetka: @cameel I did that before but you told me to fix my style.

cloud/drop-databases.yml
cameel: No, I meant the whitespace. Your code above looks like this: `{ a: b}`.
It should be either `{ a: b }` or `{a: b}` instead.

cloud/create-databases.yml
cameel: See https://github.com/golemfactory/concent-deployment/pull/226#discussion_r218452341

kubernetes/jobs/migrate-database.yml.j2
```diff
@@ -32,3 +37,19 @@ spec:
             - name:       database-job-settings
               mountPath:  /srv/http/concent_api/concent_api/settings/config_map/
               readOnly:   true
+        {% if initialize_django_admin %}
+        - name:             create-django-admin
+          image:            {{ docker_registry }}/{{ image_prefix }}concent-api:{{ image_version }}
+          imagePullPolicy:  Always
```
cameel: This script is not supposed to load the admin fixture. It's not safe because the database you're migrating may have already been initialized.

This part should be only in `reset-database`.

cloud/drop-databases.yml
bartoszbetka: Done.

cloud/create-databases.yml
bartoszbetka: Your solution repeat the `database_state` variable.
```
- {role: cloud_sql, cluster_type: control, database_state: present}
- {role: cloud_sql, cluster_type: storage, database_state: present}

File: cloud/create-databases.yml (hunk 8 above)
Comment (bartoszbetka) [reply cloud/create-databases.yml]: Done

File: containers/postgresql/reset-database.sh (hunk 10 above)
Comment (bartoszbetka) [reply containers/postgresql/reset-database.sh]: Done.

File: containers/postgresql/reset-database.sh (hunk 11 above)
Comment (bartoszbetka) [reply containers/postgresql/reset-database.sh]: Done.

File: concent-builder/configure.yml (hunk 12 above)
Comment (bartoszbetka) [reply concent-builder/configure.yml]: Done.

File: cloud/roles/postgres_connection/tasks/create-databases.yml [hunk 14]
```diff
@@ -0,0 +1,42 @@
+- name: Create the "{{ vars[ cluster_type + '_database_user'] }}" role for manage "{{ vars[ cluster_type + '_database_name'] }}" database
+  postgresql_user:
+    login_host:      "{{ control_database_host }}"
+    login_password:  "{{ control_db_admin_password | b64decode }}"
+    name:            "{{ vars[ cluster_type + '_database_user'] }}"
+    password:        "{{ vars[ cluster_type + '_db_user_password'] | b64decode }}"
+    role_attr_flags: CREATEROLE
+
+- name: Create the "{{ vars[ cluster_type + '_database_name'] }}" database
```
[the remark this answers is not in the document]
bartoszbetka: Done.

cloud/roles/postgres_connection/tasks/create-databases.yml
```diff
@@ -0,0 +1,42 @@
...
+  postgresql_privs:
+    login_host:      "{{ control_database_host }}"
+    login_password:  "{{ control_db_admin_password | b64decode }}"
+    db:              "{{ vars[ cluster_type + '_database_name'] }}"
+    role:            "{{ vars[ cluster_type + '_database_user'] }}"
+    type:            database
+    objs:            "{{ vars[ cluster_type + '_database_name'] }}"
+    privs:           ALL
+
+- name: Add postgres to role "{{ vars[ cluster_type + '_database_user'] }}"
```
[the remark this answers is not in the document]
bartoszbetka: Done.

cloud/roles/postgres_connection/tasks/create-databases.yml
```diff
@@ -0,0 +1,42 @@
...
+    role_attr_flags: CREATEROLE
+
+- name: Create the "{{ vars[ cluster_type + '_database_name'] }}" database
+  postgresql_db:
+    login_host:      "{{ control_database_host }}"
+    login_password:  "{{ control_db_admin_password | b64decode }}"
+    name:            "{{ vars[ cluster_type + '_database_name'] }}"
+    owner:           postgres
+
+- name: Add privileges for the "{{ vars[ cluster_type + '_database_user'] }}" role to the "{{ vars[ cluster_type + '_database_name'] }}" database
```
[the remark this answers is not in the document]
bartoszbetka: Done.

cloud/roles/postgres_connection/tasks/create-databases.yml
```diff
@@ -0,0 +1,42 @@
+- name: Create the "{{ vars[ cluster_type + '_database_user'] }}" role for manage "{{ vars[ cluster_type + '_database_name'] }}" database
```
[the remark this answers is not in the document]
bartoszbetka: Done.

cloud/roles/postgres_connection/tasks/delete-databases.yml
```diff
@@ -0,0 +1,17 @@
+- name: Delete the "{{ vars[ cluster_type + '_database_name'] }}" database
+  postgresql_db:
+    login_host:      "{{ control_database_host }}"
+    login_password:  "{{ control_db_admin_password | b64decode }}"
+    state:           absent
+    name:            "{{ vars[ cluster_type + '_database_name'] }}"
+    owner:           postgres
+
+- name: Delete the "{{ vars[ cluster_type + '_database_user'] }}" role for manage "{{ vars[ cluster_type + '_database_name'] }}" database
```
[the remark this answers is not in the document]
bartoszbetka: Done.

cloud/roles/postgres_connection/tasks/create-databases.yml
```diff
@@ -0,0 +1,42 @@
...
+  postgresql_user:
+    login_host:      "{{ control_database_host }}"
+    login_password:  "{{ control_db_admin_password | b64decode }}"
```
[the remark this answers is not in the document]
bartoszbetka: The `login_host` variable is the same for storage and control database. The problem is when a `storage_database_host: "{{ control_database_host }}"`. The ansible use variable as string `"{{ control_database_host }}"` which cause a error. 
```
unable to connect to database: could not translate host name "{{ control_database_host }}" to address: Name or service not known
```
We must modify https://github.com/golemfactory/concent-deployment-values/blob/master/var.yml#L38 and put the same value as for `control_database_host`.

cloud/roles/postgres_connection/tasks/create-databases.yml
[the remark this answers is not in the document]
bartoszbetka: Done.

containers/postgresql/reset-database.sh
```diff
@@ -0,0 +1,45 @@
...
+    exit 1
+fi
+
+
+schema_names_stripped=${schema_names// }
+
+# Create an array containing lines from $schema_count_stripped.
```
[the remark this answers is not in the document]
bartoszbetka: Done.

kubernetes/jobs/migrate-database.yml.j2
cameel: If you delete it and create it from scratch, you should run `reset-database` to initialize it. `migrate` is only for migrating an already initialized database.

> When database already have to admin fixture, it is not loaded again.

But it will report an error which we do not want.

cameel: If there's a case where you need to create a database without doing a reset, you need a separate job that initializes it (migrates it and loads fixtures).

concent-builder/migrate-db.yml
```diff
@@ -20,7 +20,7 @@
             chdir: "{{ build_dir}}/concent-deployment/kubernetes/build/"
 
         - name:    Create up to date secrets and config maps required by jobs
-          command: "{{ build_dir }}/concent-deployment/kubernetes/build/create-job-dependencies.sh {{ cluster_type }}"
+          command: "{{ build_dir }}/concent-deployment/kubernetes/build/create-job-dependencies.sh"
```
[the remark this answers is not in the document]
cameel: You can have the config map required by the job always on the cluster.

kubernetes/jobs/migrate-database.yml.j2
bartoszbetka: Ok. I delete load fixture from `migrate-database.yml` playbook. The solution you offered will be considered in other pull request.

README.md
```diff
@@ -157,38 +157,55 @@ ansible-playbook build-test-and-push.yml                           \
     --user       $user
 ```
 
-### Creating the database
```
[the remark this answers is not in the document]

PR #235, opened 2018-09-14
Add separate secret deployment from cluster deployment

### LINKED ISSUE #204: Separate secret deployment from cluster deployment
This task depends on #203.

### Overview
Currently we destroy and recreate secrets as a part of the cluster deployment process. This means that you need to be able to access the secrets to put a new version on the cluster.

We'd like to change that so that secrets are uploaded to the cluster once and then every subsequent deployment uses these secrets. Secrets will now change only we want to actually change passwords or certificates, not all the time.

This is a step towards encrypting secrets at rest. Such encryption will require using a password or a key every time secrets are deployed so deploying them with every change to the code would quickly become a burden, especially on the dev clusters. We could treat dev differently but we don't really have to if we introduce the mechanism described here.

### Playbooks
In addition to the stuff below, remove `var-secrets.yml` from any playbooks that don't really need them.

#### `cluster-deploy-secrets.yml`
This is a new playbook. It's meant to be run when secrets need to be uploaded to a clean cluster or updated on an existing cluster.

Steps:
- Builds secrets using `secrets/Makefile`.
- Deploys secrets to a cluster using `secrets/apply-secrets.sh`.
- Cleans the `build/` directory so that unencrypted secrets are not left laying around.
    - This should be done even if the operation fails.

#### `cluster-remove-secrets.yml`
This is a new playbook. It's meant to be run when we want to completely remove secrets from the cluster. It's only needed when we want to stop using the application on a given cluster or when new secrets no longer contain some files and we don't want to leave them on the cluster. To simply update older secrets use `deploy-secrets.yml` instead.

Steps:
- Deletes all secrets from the `default` namespace on the cluster.
    - **All** secrets - not only the ones created by `apply-secrets.sh`.

#### `build-test-and-push.yml`
- Remove the step that builds secrets.

#### `deploy.sh`
- Remove the step that deploys the secrets to the cluster.

#### `install-repositories.yml`
No changes here (yet). This playbook is still responsible for uploading secrets to the cluster. All other playbooks should assume that the secrets are already on the server.

### README
Update deployment instructions in the README. Deploying/deleting secrets and creating/deleting a database (#198) should now be in a separate section called "Cloud Management", separate from the one that describes cluster deployment.

bartoszbetka:
Resolves #204

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Do this if you want to use the remote server for building and deploying.
 
 All the instructions below assume that you're using the remote server.
 
-### Uploading secrets, building containers and cluster configuration
+### Building containers and cluster configuration
 
 ``` bash
 cd concent-deployment/concent-builder/
@@ -168,6 +168,20 @@ ansible-playbook build-test-and-push.yml                           \
     --inventory  ../../concent-deployment-values/ansible_inventory \
     --user       $user
 ```
+
+### Deploying secrets
+Before you can deploy containers, you need to make sure that certificates, keys and passwords used to configure those containers are available on the cluster.
+Deploy them with:
+
+``` bash
+cd concent-deployment/cloud/
+ansible-playbook cluster-deploy-secrets.yml                        \
+    --extra-vars cluster=$cluster                                  \
+    --inventory  ../../concent-deployment-values/ansible_inventory \
+    --user       $user
+```
+This step is necessary only when deploying the cluster for the first time or when the secrets change.
+Secret deployment is separate from deployment of the application specifically so that they can be performed separately, possibly from different machines.
 
 ### Deploying to the cluster
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,4 +1,5 @@
 ### NEXT
+- Separated secret deployment from cluster deployment (#204).
 - The concent-builder server now only uses python 3 (#249).
 - Separated cloud management from cluster deployment (#226).
 - Added possibility to use an external Signing Service (#246).
--- a/cloud/cluster-deploy-secrets.yml
+++ b/cloud/cluster-deploy-secrets.yml
@@ -1,0 +1,33 @@
+- hosts:
+    - concent-builder
+  vars_files:
+    - ../concent-builder/consts.yml
+    - "{{ deployment_values }}/var.yml"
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -167,6 +167,32 @@ ansible-playbook deploy.yml                                        \
     --user       $user
 ```
 
+### Migrate or reset the database
+
+Control and storage clusters have separate databases that need to be migrated individually.
+In some case, we can also reset database to clear all table information.
```
cameel: This should not be in this pull request. It's probably coming from the branch below. Please rebase.

README.md
```diff
@@ -184,30 +210,24 @@ ansible-playbook drop-databases.yml                                \
     --user       $user
 ```
 
-### Migrate or reset the database
+### Deploy or delete the secrets
+This is a step that allows to modify kubernetes secrets.
+In normal scenario, we don't needed to use them, only when secrets are updated or clear all component on the cluster.
```

```` markdown
### Deploying secrets
Before you can deploy containers, you need to make sure that certificates, keys and passwords used to configure those containers are available on the cluster.
Deploy them with:

``` bash
cd concent-deployment/cloud/
ansible-playbook cluster-deploy-secrets.yml                        \
    --inventory  ../../concent-deployment-values/ansible_inventory \
    --user       $user
```
This step is necessary only when deploying the cluster for the first time or when the secrets change.
Secret deployment is separate from deployment of the application specifically so that they can be performed separately, possibly from different machines.
````

Info about removing is not necessary. It's not a part of the usual workflow.

BTW, please keep a space between ```` ``` ```` and `bash`.

Comment (cameel) [inline README.md]: Please put this section after `Uploading secrets, building containers and cluster configuration`.

And rename `Uploading secrets, building containers and cluster configuration` to `Building containers and cluster configuration`.

File: README.md [hunk 3]
```diff
@@ -184,30 +210,24 @@ ansible-playbook drop-databases.yml                                \
...
-
-ansible-playbook migrate-db.yml                                    \
-    --extra-vars "cluster=$cluster cluster_type=$cluster_type"     \
+ansible-playbook cluster-remove-secrets.yml                        \
+    --extra-vars cluster=$cluster                                  \
     --inventory  ../../concent-deployment-values/ansible_inventory \
     --user       $user
 ```
 
+
```
cameel: This line is not ncecessary.

README.md

README.md

PR #238, opened 2018-09-17
Add `middleman_protocol` app to concent container

bartoszbetka:


containers/concent-api/install-concent-api.sh.j2
```diff
@@ -8,3 +8,7 @@ source "$virtualenv_dir/bin/activate"
 python -m pip install --upgrade pip
 python -m pip install --requirement "/srv/http/concent_api/requirements.lock"
 python -m pip install gunicorn=="{{ gunicorn_version }}"
+(
+    cd "/srv/http/middleman_protocol/"
+    python setup.py develop
+)
```
cameel: Offtopic: we should be clearing pip cache after installation in this script. You could do it in a separate pull request.

cameel: Please create a separate script. We don't want to have to reinstall other dependencies when the source changes.

containers/package-builder/build-virtualenv.sh.j2
```diff
@@ -8,3 +8,7 @@ mkdir --parents "$build_dir/virtualenv/"
 virtualenv --python "python{{ python_version }}" "$build_dir/virtualenv/"
 source "$build_dir/virtualenv/bin/activate"
 python -m pip install -r "$requirements_path"
+(
+    cd "$build_dir/repositories/concent/middleman_protocol/"
+    python setup.py develop
+)
```
cameel: Same as above. Put it in a separate script so that you don't have to change the order in the `Dockerfile`.

containers/concent-api/install-concent-api.sh.j2
bartoszbetka: Ok. I will make it in separate pull request.

bartoszbetka: Done

containers/concent-api/install-concent-api.sh.j2
```diff
@@ -6,6 +6,7 @@ virtualenv_dir=/srv/http/virtualenv
 virtualenv --python "python{{ python_version }}" "$virtualenv_dir"
 source "$virtualenv_dir/bin/activate"
 python -m pip install --upgrade pip
+cd /srv/http/concent_api/
```
cameel: Please add a comment that requirements needs to be installed from the directory it's in because it contains relative paths.

containers/Makefile
```diff
@@ -110,20 +110,20 @@ build/concent-api/requirements.lock: build/concent-api/concent-api.tar
 build/concent-api/testing.pref: ../concent-builder/files/testing.pref
 	cp "$<" "$@"
 
-build/concent-api/concent-api.tar: build/repositories/concent/  build/virtualenv/
+build/concent-api/concent-api.tar: build/repositories/concent  build/virtualenv/
```
cameel: Cosmetic changes like this should be in a separate commit, especially when there's more of them like in this case.

containers/concent-api/install-concent-api.sh.j2
bartoszbetka: Done.

containers/Makefile

PR #246, opened 2018-10-02
Add possibility to use concent and middleman with external signing-service

bartoszbetka:


kubernetes/config-maps/concent-api-worker/local_settings.py.j2
```diff
@@ -18,7 +18,7 @@ DEBUG_INFO_IN_ERROR_RESPONSES = {{ allow_debug_info_in_error_responses }}
 
 LOGGING['handlers']['console']['level'] = DEBUG
 
-{% if signing_service_on_the_cluster %}
+{% if allow_signing_service_authentication %}
 USE_SIGNING_SERVICE = True
 {% endif %}
```
cameel: When `USE_SIGNING_SERVICE` is `False`, the configuration must contain the private key. Otherwise Concent won't be able to sign transactions.

cameel: You should set this to `False` explicitly instead of relying on a default.

Something like:
``` python
USE_SIGNING_SERVICE = {{ True if allow_signing_service_authentication else False }}
```
(not tested; you may have to use `"True"` instead of `True`).

kubernetes/create-services.sh.j2
```diff
@@ -8,7 +8,9 @@ kubectl create --record --filename services/verifier.yml
 {% endif %}
 kubectl create --record --filename services/geth.yml
 kubectl create --record --filename services/rabbitmq.yml
+{% if allow_signing_service_authentication %}
 kubectl create --record --filename services/middleman.yml
+{% endif %}
 {% if signing_service_on_the_cluster %}
```
cameel: When not deploying middleman, you should also disable the TCP proxy in nginx.

kubernetes/config-maps/concent-api-worker/local_settings.py.j2
bartoszbetka: Done.

bartoszbetka: Done.

kubernetes/create-services.sh.j2
bartoszbetka: Done.

PR #249, opened 2018-10-03
Chore stop using python2 on the concent builder server

### LINKED ISSUE #227: Stop using Python 2 on concent-builder
Looks like both Ansible and yasha now support Python 3:
- [Ansible > Python 3 Support](https://docs.ansible.com/ansible/2.6/reference_appendices/python_3_support.html)
- [yasha README](https://github.com/kblomqvist/yasha/blob/master/README.md)

So we finally stop using Python 2 on our machines.

Update concent-builder playbooks so that Python 2 is no longer installed or used for anything.

If we're using Python 2 anywhere else in our deployment, leave a comment below and get rid of it too.

bartoszbetka:
Resolves #227

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,4 +1,5 @@
 ### NEXT
+- The concent-builder server now only uses python 3 (#249).
 - Separated cloud management from cluster deployment (#226).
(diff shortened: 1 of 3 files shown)


cameel:
Could you select the right branch as a base? Now this includes also branches below.

PR #254, opened 2018-10-12
Add script `delete-old-files.sh` that delete old files and checking free space in the nginx-storage

bartoszbetka:


containers/nginx-storage/delete-old-files.sh
```diff
@@ -0,0 +1,39 @@
...
+DIRECTORIES="$(echo /srv/storage/blender/{result/,'source/',verifier-output/})"
+
+for directory in ${DIRECTORIES[@]}; do
+    if [ ! -d "$directory" ]; then
+        echo "[crond][error][disk cleanup]: Please check if the \"$directory\" directory exist."
+        exit 1
+    fi
+done
+
+free_space="$(df -h | grep '/srv/storage' | awk '{ print $4; }' | tr -d 'G')"
```
michalpiasecki1: does it remove other characters other than 'G' like 'M' (megabytes) ? maybe better would be to remove -h (from df), and then parse raw bytes?

containers/nginx-storage/delete-old-files.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -e
+
+DIRECTORIES="$(echo /srv/storage/blender/{result/,'source/',verifier-output/})"
```
michalpiasecki1: is it good idea to replace it with a list, and then run `find` for each element in a list ? This is hacky, but less readable. Also, [useless use of echo](http://porkmail.org/era/unix/award.html) ;)

containers/nginx-storage/delete-old-files.sh
```diff
@@ -0,0 +1,39 @@
...
+if [ -z "$free_space" ]; then
+    echo "[crond][error][disk cleanup]: There is a problem with checking free space on disk. Please check if mount point \"/srv/storage/\" exist."
+    exit 1
+fi
+
+find $DIRECTORIES      \
+    -mtime +30         \
+    -delete
+
+# Remove the part after the dot
```
michalpiasecki1: rounding down free space to full GB? MB?

containers/nginx-storage/delete-old-files.sh
```diff
@@ -0,0 +1,39 @@
...
+free_space=${free_space%.*}
+days=25
```
michalpiasecki1: If I am not wrong this will remove files older than 30 days. If it is not sufficient, then try 25,24,23...16. Why not check 29-26 as well ?

containers/nginx-storage/delete-old-files.sh
bartoszbetka: Done.

containers/nginx-storage/delete-old-files.sh
bartoszbetka: Done.

containers/nginx-storage/delete-old-files.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo "[crond][error][disk cleanup]: There is a problem with checking free space on disk. Please check if mount point \"/srv/storage/\" exist."
+    exit 1
+fi
+
+delete_old_files 30
+
+# Convert kilobyte to gigabyte
+let free_space=$free_space/1048576
+days=29
+while (( 15 > $free_space ))
```
dybi: please extract it to a variable, i. e. `$MINIMAL_FREE_SPACE_IN_GB`

containers/nginx-storage/delete-old-files.sh
```diff
@@ -0,0 +1,62 @@
...
+free_space="$(df | grep '/srv/storage' | awk '{ print $4; }')"
+if [ -z "$free_space" ]; then
+    echo "[crond][error][disk cleanup]: There is a problem with checking free space on disk. Please check if mount point \"/srv/storage/\" exist."
+    exit 1
+fi
+
+delete_old_files 30
+
+# Convert kilobyte to gigabyte
+let free_space=$free_space/1048576
```
dybi: free space doesn't get recalculated after deletion?

containers/nginx-storage/delete-old-files.sh
```diff
@@ -0,0 +1,62 @@
...
+    fi
+done
+
+free_space="$(df | grep '/srv/storage' | awk '{ print $4; }')"
+if [ -z "$free_space" ]; then
+    echo "[crond][error][disk cleanup]: There is a problem with checking free space on disk. Please check if mount point \"/srv/storage/\" exist."
+    exit 1
+fi
+
+delete_old_files 30
```
dybi: please extract days to a variable and maybe make it a script parameter

containers/nginx-storage/delete-old-files.sh
bartoszbetka: Done.

containers/nginx-storage/crontab.txt
```diff
@@ -0,0 +1,2 @@
+# min   hour    day     month   weekday command
+0       2       *       *       *       /etc/crontabs/delete-old-files.sh > /dev/stdout
```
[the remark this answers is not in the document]
bartoszbetka: Our errors go to stdout, but the commands errors go to stderr, so it's good idea to do this.

containers/nginx-storage/delete-old-files.sh
bartoszbetka: You're right, thanks.

containers/nginx-storage/crontab.txt
```diff
@@ -1,2 +1,2 @@
 # min   hour    day     month   weekday command
-0       2       *       *       *       /etc/crontabs/delete-old-files.sh > /dev/stdout
+0       2       *       *       *       /etc/crontabs/delete-old-files.sh 2> /dev/stderr > /dev/stdout
```
dybi: why not just? 
```bash
/etc/crontabs/delete-old-files.sh > /dev/stdout  2>&1
```
?

containers/nginx-storage/delete-old-files.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash -e
 
+MINIMAL_FREE_SPACE_IN_GB=15
+# Variable that control deleting files older than specify value.
+MAXIMUM_DAYS=30
+# Variable that control lower limit for deleting files older than specify value,
```
dybi: specify-> specified

containers/nginx-storage/delete-old-files.sh
```diff
@@ -28,34 +37,39 @@ function delete_empty_directories {
...
-
-delete_old_files 30
+available_free_space=$(check_free_space)
+delete_old_files $MAXIMUM_DAYS
 
-# Convert kilobyte to gigabyte
-let free_space=$free_space/1048576
-days=29
-while (( 15 > $free_space ))
+let MAXIMUM_DAYS--
```
dybi: I would add third variable: 
```
let CURRENT_NUMBER_OF_DAYS_TO_DELETE=$MAXIMUM_DAYS-1
```
or sth similar (even `DAYS_TO_DELETE`)
and then use it in calculations, not to modify `MAXIMUM_DAYS`

containers/nginx-storage/crontab.txt
bartoszbetka: Because it redirects `stderr` to `stdout`.

containers/nginx-storage/delete-old-files.sh
bartoszbetka: Done.

containers/nginx-storage/delete-old-files.sh
bartoszbetka: Done.

PR #260, opened 2018-11-06
Chore switching between kubernetes clusters without modifying global settings

### LINKED ISSUE #242: Switching between Kubernetes without modifying global settings
Currently all our Ansible playbooks use `gcloud container clusters get-credentials` command to switch to the right cluster before running scripts that use `kubectl`:

``` yaml
- name:    Configure kubectl to operate on the right cluster
  command: gcloud container clusters get-credentials \
      "{{ gke.cluster }}"                            \
      --project "{{ gke.project }}"                  \
      --zone    "{{ gke.zone }}"
```

This works but only as long as we don't run more than one playbook at a time. That's because this command saves the choice permanently in gcloud's configuration file in user's home directory. If two playbooks run simultaneously, choice made by one of them will be overwritten by the other. This also interferes with using `kubectl` in another terminal while a playbook is running in the background.

Another bad side effect is that when using `kubectl` manually, it's easy to forget to switch the cluster and make changes to the wrong one.

To fix this we need to stop using `gcloud container clusters get-credentials` and instead create separate `kubeconfig` files for `kubectl` when the build server is configured. Then pass the config to each `kubectl` call explicitly.

Also, preferably there should be no cluster selected by default so that the command fails if you forget to specify the config.

As for `gcloud` commands, there must be a way to have them take cluster name as a parameter.

bartoszbetka:
Resolves #242


concent-builder/configure.yml
```diff
@@ -111,3 +111,10 @@
             state: directory
             owner: "{{ shared_user }}"
             group: "{{ shared_user }}"
+
+        - name: Upload configuration for kubectl
+          copy:
+            src:   "{{ local_secret_dir }}/cloud/config"
```
cameel: Please call the file `kubeconfig` to make it clear what it is.

concent-builder/configure.yml
```diff
@@ -111,3 +111,10 @@
...
+            dest:  "/home/{{ shared_user }}/.kube/config"
+            owner: "{{ shared_user }}"
+            group: "{{ shared_user }}"
```
cameel: One problem I see: we're doing this only for `shared_user`. Don't we need to do this on the account of every user who might run the playbook? Or is everyone still deploying from this user account?

concent-builder/configure.yml
cameel: We need some information in the README on how to update this file when it expires or when the IP changes (I see that the IP is hard-coded inside).

concent-builder/configure.yml
```diff
@@ -111,3 +111,10 @@
             state: directory
             owner: "{{ shared_user }}"
             group: "{{ shared_user }}"
+
+        - name: Upload configuration for kubectl
+          copy:
+            src:   "{{ local_secret_dir }}/cloud/config"
+            dest:  "/home/{{ shared_user }}/.kube/config"
```
cameel: Replacing the whole `~/.kube/config` has basically the same effect as `gcloud container clusters get-credentials`. It still replaces the configuration globally and does not let us manage multiple clusters in parallel which was one of the problems listed in the issue.

We need a better solution. Is it possible to have several configurations and specify them by path in `kubectl`? Or maybe we could install the configurations in locations like `~/.kube/config.d/concent-dev`, `~/.kube/config.d/concent-staging`, etc and then add a step that generates `~/.kube/config` by merging them?

concent-builder/configure.yml
bartoszbetka: Yes with still use one user account for deploying. https://github.com/golemfactory/concent-deployment/pull/223

concent-builder/configure.yml
bartoszbetka: This file use `gcloudsdk` to update credentials. This is original config, only context names are changes.

concent-builder/configure.yml
bartoszbetka: We don't need a few configuration file. Kubectl allow to specify context by `--context=name` like for example `concent-dev`. This solution allow manage multiple clusters in parallel.

concent-builder/configure.yml
cameel: But you said you changed some stuff in that file. We need information on how to add it to our secrets and how to update it.

concent-builder/configure.yml
cameel: After our discussion I understand that this file contains credentials for all the clusters. If we treat it as a part of the cloud secrets then it's fine. I mistook it for a part of cluster secrets that would need to change for each cluster.

PR #263, opened 2018-11-14
Update the `README.md` file after add extend signing-service with daily threshold checks

bartoszbetka:


README.md
```diff
@@ -344,22 +344,23 @@ If you have received the package and only want to build and run the docker conta
...
-        --sentry-dsn-from-env                               \
+docker run                                                                                         \
+    --detach                                                                                       \
+    --env         ETHEREUM_PRIVATE_KEY                                                             \
+    --env         SIGNING_SERVICE_PRIVATE_KEY                                                      \
+    --env         SENTRY_DSN                                                                       \
+    --network     host                                                                             \
+    --hostname    signing-service                                                                  \
+    --name        signing-service                                                                  \
+    --volume      /var/daily_thresholds:/usr/lib/signing_service/signing-service/daily_thresholds  \
```
cameel: I think that `/var/log/concent/daily_thresholds/` might be a better directory for those files on the host. In the container too, actually, but that would require a change in concent code. The application is not supposed to have write access to its directory in `/usr/lib/`

README.md
```diff
@@ -374,6 +375,8 @@ This assumes that:
     Otherwise just skip the `--sentry-dsn-from-env` parameter and the `SENTRY_DSN` variable.
 - `mainnet` is the name of the Sentry environment that should be included in error reports.
 
+The volume is needed as persistent disc for daily thresholds report.
+The owner of mount point directory on the system side must be set to uid 999 which is id of running user on the docker.
```
cameel: Please remove this paragraph and add this to the bullet list above instead:
``` markdown
- `/var/daily_thresholds/` is location in the host system where the Signing Service can create daily threshold reports.
    It must be writable by a user with UID of 999 which is the UID under which the application runs in the container.
```

README.md
bartoszbetka: The application is not have access to write in different directory then `/usr/lib/signing_service/signing-service`. All other files and directories are own by root.

bartoszbetka: The path was changed on the host side.

PR #273, opened 2018-11-26
Connect to an external geth node instead of setting up one inside the machine

### LINKED ISSUE #268: Concent VM: Connect to an external Geth node instead of setting up one inside the machine
Geth requires a lot of space and is a common point of failure for us. I think that switching to an external node will make things easier for people using the virtual machine (#147).

- Remove geth from the machine.
- Remove extra disk images created to store the blockchain.
- Configure Golem and Concent to use an external node.

bartoszbetka:
Resolves #268

### DIFF
--- a/concent-vm/Vagrantfile
+++ b/concent-vm/Vagrantfile
@@ -12,24 +12,15 @@ Vagrant.configure("2") do |config|
 
     config.vm.network "private_network", ip: "172.40.2.3"
 
-    blockchain_disk       = './disk/blockchain_disk.vdi'
-    golem_blockchain_disk = './disk/golem_blockchain_disk.vdi'
+    golem_configuration_directory = './disk/golem_configuration_directory.vdi'
 
     config.vm.provider "virtualbox" do |virtualbox|
         virtualbox.memory = 2048
         virtualbox.gui    = false
-        if not File.exists?(blockchain_disk)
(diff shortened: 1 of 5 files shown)


concent-vm/Vagrantfile
```diff
@@ -12,24 +12,15 @@ Vagrant.configure("2") do |config|
 
     config.vm.network "private_network", ip: "172.40.2.3"
 
-    blockchain_disk       = './disk/blockchain_disk.vdi'
-    golem_blockchain_disk = './disk/golem_blockchain_disk.vdi'
+    golem_configuration_directory = './disk/golem_configuration_directory.vdi'
```
cameel: Instead of creating a separate disk, it would be better to create a directory on the host and share it with the machine. The reason why we were using a disk was that sharing a directory is not the best idea if you have lots of data because by default it gets copied into the machine with `rsync`. But Golem's configuration is not big.

Please use [`vboxsf`](https://help.ubuntu.com/community/VirtualBox/SharedFolders).

It's fine to do this in a separate pull request.

bartoszbetka: Ok. This will be doing in a separate pull request.

cameel: I've created #309 for that.

PR #275, opened 2018-11-30
Add the missing package source for the gcloud and kubectl packages

bartoszbetka:


bartoszbetka:
This code existed earlier. We deleted its accidentally, when we merged the code of virtual machine.

bartoszbetka:
This code existed earlier. We deleted its accidentally, when we merged the code of virtual machine.

cameel:
OK. Please always add information like this in the commits (and may also in the pull request). It's important when you need to decide if the change makes sense.

PR #277, opened 2018-12-07
Add helper script for deployment

bartoszbetka:
This is a helper scripts for deployment process. The `delete-database.sh` script depend on #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ ansible-playbook install-secrets.yml                               \
     --user       $user
 ```
 
-All the instructions below assume that you're using local playbooks to run build and deployment commands on the remote server.
+All the instructions below assume that you're using local playbooks to run build and deployment commands on a new server.
 
 Note that if you're running the playbooks themselves from within that server too, you need to add `--connection=local` to your `ansible-playbook` calls.
 Otherwise Ansible will run its commands over SSH (rather than directly) using the public IP specified in `ansible_inventory`, which will likely fail because you're not supposed to have your private SSH key on the remote server you're connecting to.
@@ -271,6 +271,32 @@ ansible-playbook migrate-db.yml                                    \
 ```
 
 It's safe to run migrations even if there are no changes - Django will detect that and simply leave the schema as is.
+
+#### Helper script for deployment process
+
+###### `build-and-deploy-to-dev.sh`
+This script provides a streamlined way to deploy a development version of the code to the `concent-dev` cluster.
+The use case could be automated because it does not require as much flexibility as the usual deployment procedure.
+The `concent-dev` cluster does not hold any production data and when something goes wrong it's fine to just remove everything and start from scratch.
+
+The `configure.yml` playbook installs the script on the build server where it can use secrets necessary to access the cluster.
+Then it can be run from that server by any user.
+
+The script deploys the code from the `dev` branch in the `concent-deployment` repository.
+Version of the `concent` repository is specified in `concent_version` in `containers/versions.yml`.
+To deploy any other version you must change this value in your local working copy, push the commit to the repository and update the `dev` branch to point at it.
+The value can be a commit ID, tag, branch.
+
+The script performs its job by running the Ansible playbooks listed above in the right order and with the parameters suitable for the `concent-dev` cluster.
+
+Command-line options:
+1. **database_operation**: _migrate_ (default) or _reset_.
+    _migrate_ does not clear the database runs the migrations.
+    _reset_ clears the database and reinitializes it.
+2. **deployment_dir**: the directory to store a clone of the `concent-deployment` repository.
+    `~/deployment` by default.
+    This working copy should not be modified manually.
+    The script will recreate it if it does not exist or update it if it does but it will fail if a checkout is not possible (e.g. because there are uncommitted changes).
 
 ### Building nginx-storage locally
 
--- a/kubernetes/clean-up-jobs.sh
+++ b/kubernetes/clean-up-jobs.sh
@@ -1,4 +1,6 @@
 #!/bin/bash -e
 
-kubectl delete job reset-database   || true
-kubectl delete job migrate-database || true
+kubectl delete job reset-control-database   || true
+kubectl delete job reset-storage-database   || true
+kubectl delete job migrate-control-database || true
+kubectl delete job migrate-storage-database || true
--- a/concent-builder/configure.yml
+++ b/concent-builder/configure.yml
@@ -121,3 +121,8 @@
             name:    nginx
             state:   restarted
 
+        - name:   Add helper script for deployment
+          copy:
+            src:  build-and-deploy-to-dev.sh
+            dest: /usr/local/bin/build-and-deploy-to-dev.sh
+            mode: 0555
--- a/concent-builder/files/build-and-deploy-to-dev.sh
+++ b/concent-builder/files/build-and-deploy-to-dev.sh
@@ -1,0 +1,78 @@
+#! /bin/bash -e
+
+export ANSIBLE_STDOUT_CALLBACK=debug
+DEFAULT_DATABASE_OPERATION=migrate
+DEFAULT_DEPLOYMENT_DIR=~/deployment
+database_operation="${1:-$DEFAULT_DATABASE_OPERATION}"
+deployment_dir="${2:-$DEFAULT_DEPLOYMENT_DIR}"
+
+function ansible_command {
+    echo "playbook: '$1', cluster: '$2', cluster type: '$3'"
+    file="$1"
+    cluster_name="$2"
+    cluster_type="$3"
+    ansible-playbook                                                      \
+        $file                                                             \
+        --connection local                                                \
+        --extra-vars "cluster=$cluster_name cluster_type=$cluster_type"   \
+        --inventory ../../concent-deployment-values/ansible_inventory
+}
+
+declare -A repositories
+
+repositories=(
+    ["concent-deployment"]="https://github.com/golemfactory/concent-deployment.git"
+    ["concent-deployment-values"]="https://github.com/golemfactory/concent-deployment-values.git"
+)
+
+if [ ! -d $deployment_dir ]; then
+    mkdir $deployment_dir/
+fi
+for repository in "${!repositories[@]}"; do
+    repository_path="$deployment_dir/$repository/"
+    if [ ! -d $repository_path ]; then
+        git clone "${repositories[$repository]}" "$repository_path"
+    else
+        cd "$repository_path"
+        git fetch origin --prune
+    fi
+done
+
+(
+    for repository in "${!repositories[@]}"; do
+        cd $deployment_dir/$repository/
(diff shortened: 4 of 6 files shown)


deploy-help/configure.sh
```diff
@@ -0,0 +1,48 @@
...
+for secrets_directory in "${!concent_dev_secrets_directories[@]}"; do
+    if [ ! -d ${concent_dev_secrets_directories[$secrets_directory]} ]; then
+        cp                                                           \
+            --recursive                                              \
+            "/var/concent-data/concent-secrets/$secrets_directory/"  \
+            "${concent_dev_secrets_directories[$secrets_directory]}"
+    fi
+done
+
+echo "concent-builder ansible_connection=local" > $deployment_dir/inventory
```
cameel: I'd just use `ansible_inventory` from concent-deployment-values and add `--connection=local` as a command-line parameter when running `ansible-playbook`. The less stateful configuration the better.

deploy-help/configure.sh
```diff
@@ -0,0 +1,48 @@
...
+echo "export ANSIBLE_STDOUT_CALLBACK=debug" >> ~/.bashrc
```
cameel: You could just `export ANSIBLE_STDOUT_CALLBACK=debug` in the scripts that run the playbooks.

deploy-help/delete-database.sh
```diff
@@ -0,0 +1,43 @@
...
+function run_kubernetes_script {
+    executed_script="$1"
+    cd $kubernetes_build_dir/
+    ./$executed_script
+}
+
+if [ -z "$cluster_name" ]; then
+    echo 'Please specify the name of database'
+    exit 1
+fi
```
cameel: I'd just make it reset both databases. I can't think of many situation on the dev cluster where you would want to delete one but keep the other. It can lead to weird, inconsistent behavior.

deploy-help/delete-database.sh
```diff
@@ -0,0 +1,43 @@
...
+echo
+if [[ $REPLY =~ ^[Yy]$ ]]; then
+
+    playbooks=(
+        job-cleanup.yml
+        reset-db.yml
+    )
+
+    run_kubernetes_script delete-services.sh
+    (
```
cameel: You should not need really need to call the deployment scripts directly here. Playbooks should be able to do this for you. I have created an issue to split `deploy.yml` into two playbooks: #281. When it's done, update this script to use it.

deploy-help/delete-database.sh
```diff
@@ -0,0 +1,43 @@
...
+    ./$executed_script
+}
+
+if [ -z "$cluster_name" ]; then
+    echo 'Please specify the name of database'
+    exit 1
+fi
+
+# Confirm executing script
+read -p "Warning must execute after deploy.sh script. Are you sure about execute this script? It's destroy database[y/n]" -n 1 -r
```
cameel: Please reword:

``` bash
echo "This script won't work correctly unless you run `deploy.sh` first because it depends on containers built by it."
read -p "This script removes everything from the databases on the dev cluster. Are you sure you want to continue?[y/N]" -n 1 -r
```

deploy-help/delete-database.sh
```diff
@@ -0,0 +1,43 @@
+#! /bin/bash -e
+
```
cameel: I think this script should be merged with `deploy.sh`. `deploy.sh` should have an option that tells it whether to migrate or destroy the database before deployment. Migration should be the default.

deploy-help/configure.sh
```diff
@@ -0,0 +1,48 @@
+#! /bin/bash -e
+
```
cameel: I think the `configure.sh` script is not necessary. It should be merged with `deploy.sh`.

deploy-help/deploy.sh
```diff
@@ -0,0 +1,41 @@
+#! /bin/bash -e
+
+deployment_dir=~/deployment
```
cameel: Please make it a parameter. Make it optional and set it to `~/deployment/` by default.

deploy-help/delete-database.sh
cameel: Actually, when you merge this script with `deploy.sh`, this question will not be necessary.

deploy-help/deploy.sh
```diff
@@ -0,0 +1,41 @@
+#! /bin/bash -e
+
```
cameel: Please rename the script to `build-and-deploy-to-dev.sh`. Put it in `concent-builder/files/` instead and make `configure.yml` upload it to `/usr/local/bin` on the server along with other scripts.

deploy-help/deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+    git checkout origin/dev
+
+    databases=(
+        control
+        storage
+    )
+    playbooks=(
+        install-repositories.yml
+        build-test-and-push.yml
+        job-cleanup.yml
```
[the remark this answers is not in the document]
bartoszbetka: Ok. I change order of playbooks and put the `job-cleanup.yml` playbook as first playbook. 

We need two times execute this playbook. The first running is to make sure that kubernetes job with migrate-db isn't running on the cluster. The second running is to cleanup after first database is migrate, to allow migrate of second database.

deploy-help/configure.sh
```diff
@@ -0,0 +1,48 @@
...
+if [ ! -d $deployment_dir/concent-secrets/ ]; then
+    mkdir $deployment_dir/concent-secrets/
+fi
+for secrets_directory in "${!concent_dev_secrets_directories[@]}"; do
+    if [ ! -d ${concent_dev_secrets_directories[$secrets_directory]} ]; then
+        cp                                                           \
+            --recursive                                              \
+            "/var/concent-data/concent-secrets/$secrets_directory/"  \
+            "${concent_dev_secrets_directories[$secrets_directory]}"
```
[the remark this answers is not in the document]
bartoszbetka: We don't need concent-secrets directory at all. We use secrets only on cloud management level.

deploy-help/deploy.sh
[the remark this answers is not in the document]
cameel: Oh, right, we have two separate migrate jobs with the same name. So let's add database name to the job name. Both these jobs are created from a template. It's just a matter of using the template variable there and then updating `clean-up-jobs.sh` to remove both jobs.

deploy-help/configure.sh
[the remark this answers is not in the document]
cameel: OK, so let's remove it.

deploy-help/deploy.sh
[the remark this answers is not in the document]
bartoszbetka: I checked and `job-cleanup.yml` can't be executed as first playbook because it's depends on `build-test-and-push.yml`.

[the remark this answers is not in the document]
cameel: True. We need to build it first. But it can still be before `deploy` rather than after it.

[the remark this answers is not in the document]
bartoszbetka: I agree with you but it's already before `deploy`.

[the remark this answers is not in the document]
cameel: Yeah, I'm not saying it's not :) I just wanted to clarify that this is what is really needed here - it does not have to be before everything (if fact it can't as you pointed out), but it has to be before `deploy` so that we can start deployment in a clean situation without worrying about job name conflicts.

README.md
```diff
@@ -258,6 +272,17 @@ ansible-playbook migrate-db.yml                                    \
 
 It's safe to run migrations even if there are no changes - Django will detect that and simply leave the schema as is.
 
+#### Helper script for deployment process
+
+###### `build-and-deploy-to-dev.sh`
+This script is available on the remote `concent-builder` server.
+Before connect to server and run script, you must locally specify version of concent in the `containers/versions.yml` file and push it as commit to dev branch on the `concent-deployment` repository.
```
cameel: Please reword this section:
``` markdown
This script provides a streamlined way to deploy a development version of the code to the `concent-dev` cluster.
The use case could be automated because it does not require as much flexibility as the usual deployment procedure.
The `concent-dev` cluster does not hold any production data and when something goes wrong it's fine to just remove everything and start from scratch.

The `configure.yml` playbook installs the script on the build server where it can use secrets necessary to access the cluster.
Then it can be run from that server by any user.

The script deploys the code from the `dev` branch in the `concent-deployment` repository.
Version of the `concent` repository is specified in `concent_version` in `containers/versions.yml`.
To deploy any other version you must change this value in your local working copy, push the commit to the repository and update the `dev` branch to point at it.
The value can be a commit ID, tag, branch.
```

README.md
```diff
@@ -258,6 +272,17 @@ ansible-playbook migrate-db.yml                                    \
...
+
+Description of the script:
+- configures environment for deployment process in your home directory. You can also specify custom path by second argument in the script.
+- builds and deploys specifics concent version on the dev cluster.
+- migrates or resets databases, default is migrate. To reset databases use `reset` as first argument in the script.
```
``` markdown
The script performs its job by running the Ansible playbooks listed above in the right order and with the parameters suitable for the `concent-dev` cluster.

Command-line options:
1. **database_operation**: _migrate_ (default) or _reset_.
    _migrate_ does not clear the database runs the migrations.
    _reset_ clears the database and reinitializes it.
2. **deployment_dir**: the directory to store a clone of the `concent-deployment` repository.
    `~/deployment` by default.
    This working copy should not be modified manually.
    The script will recreate it if it does not exist or update it if it does but it will fail if a checkout is not possible (e.g. because there are uncommitted changes).
```

README.md
```diff
@@ -152,7 +152,22 @@ Do this if you want to use the remote server for building and deploying.
 
     Where the `$user` shell variable contains the name of your shell account on the remote machine.
 
-All the instructions below assume that you're using the remote server.
+
+### Uploading secrets and access to deploy on specific gcloud environment
+This is a separate step that is needed on the new server and in the case of changes in secrets
```
cameel: `on the new server` -> `on a new server`

kubernetes/clean-up-jobs.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash -e
 
-kubectl delete job reset-database   || true
-kubectl delete job migrate-database || true
+kubectl delete job reset-database-control   || true
+kubectl delete job reset-database-storage   || true
+kubectl delete job migrate-database-control || true
+kubectl delete job migrate-database-storage || true
```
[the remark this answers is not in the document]
cameel: You changed `reset-database` to `reset-database-control` and `reset-database-storage`. Similarly for `migrate-database`. `reset-control-database` and `reset-storage-database` would be more natural and I assumed that you would do it that way.

Anyway, it's not a big deal. I've accepted the pull request anyway because your names are acceptable too. I would change them but I'll leave it up to you.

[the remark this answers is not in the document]
bartoszbetka: Done.

Issue #191, opened 2018-07-18
Git cache on build server updated on demand


cameel:
Currently we use [install-repositories.yml](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/concent-builder/install-repositories.yml) playbook as a part of the build process. It creates clones of our repositories in `/var/concent-data/` or updates them if they already exist. The problem with the playbook is that it requires root privileges so it's not usable for normal users or the build server.

The purpose of this task is to prepare a script that can be executed by normal users and updates the repositories. Creation can remain in the playbook.

#### Changes in playbooks
1. Move the repositories from `/var/concent-data/` to `/var/git-mirror/`. Update paths in the playbooks.
2. `/var/git-mirror/` should be readable (but not writable) for all users of the system.
3. Implement `update-git-mirror.sh`.
4. Remove cloning repositories from `install-repositories.yml` and rename it to `install-secrets.yml`.
5. Update [configure.yml](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/concent-builder/configure.yml) playbook:
    - It should install `update-git-mirror.sh` in `/usr/local/bin/`.
        - Use [setuid](https://www.linuxnix.com/suid-set-suid-linuxunix/) to allow all users to run this script with root privileges.
    - It should generate a list of git repositories and put it in `/var/git-mirror/repositories.sh` (see below).
        - Generate this file from a template and `repositories.yml`.
    - It should run `update-git-mirror.sh` to clone or update the repositories.
6. Update [build-test-and-push.yml](https://github.com/golemfactory/concent-deployment/blob/v0.7.3/concent-builder/build-test-and-push.yml) playbook to run `update-git-mirror.sh` before the build.

#### `repositories.sh`
The file is a bash script that defines a list with repository URLs and names and puts it in a variable. It's meant to be sourced by `update-git-mirror.sh` to load the variables.

#### `update-git-mirror.sh`
- The script should **not** let the user specify the path to the directory that stores repositories. It must be hard-coded to `/var/git-mirror/` to prevent users from overwriting important system directories.
- For each repository:
    - If it's on the repository list but not in `/var/git-mirror/`, do `git clone --mirror`.
    - If it's on the repository list and is present in `/var/git-mirror/`, update it with `git fetch --prune`.
    - If it's not on the repository list but is present in `/var/git-mirror/`, remove it from `/var/git-mirror/`.
- All files in `/var/git-mirror/` should be owned by root and readable to other users.
[...]

PR #284, opened 2018-12-14
Add jenkins ci to concent builder server

### LINKED ISSUE #297: Jenkins on build server as a CI server for Concent
Create a playbook that installs Jenkins on our build server and configures it as CI for Concent.

Requirements:
- Stateless - it should not require backing up user accounts. We should be able to recreate everything from scratch by running the playbook
    - Secrets should be stored in the same place where we store cloud and cluster secrets
- Read-only - users should only be able to view build results. All changes to the build pipeline should be done by modifying the configuration in `concent-deployment` and `concent-deployment-values` repositories.
- No polling - use webhooks on github to notify jenkins that the code must be rebuilt.

Configure Jenkins to watch the branch called `ci` in `concent-deployment` repository and run the build whenever anything in it changes.

bartoszbetka:
Resolves #297.
It's depend on #51 
When the pull request will approve, the address of concent-deployment repository must be change to originally, not forked and the webhook on that repository must be configured.


concent-builder/jenkins-configuration-files/config.xml
```diff
@@ -0,0 +1,84 @@
...
+          <spec></spec>
+        </com.cloudbees.jenkins.GitHubPushTrigger>
+      </triggers>
+    </org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty>
+  </properties>
+  <definition class="org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition" plugin="workflow-cps@2.61">
+    <script>properties([[$class:
+&apos;GithubProjectProperty&apos;,
+    displayName: &apos;&apos;,
+    projectUrlStr: &apos;https://github.com/bartoszbetka/concent-deployment.git/&apos;],
```
cameel: This URL should be in our configuration in concent-deployment-values. It might be a good idea to have a separate configuration file just for concent-builder or even just for Jenkins. We don't want URLs hard-coded in the code repository.

cameel: Please remember to replace this URL with the address of the main concent-deployment repository.

concent-builder/jenkins-configuration-files/config.xml
```diff
@@ -0,0 +1,84 @@
...
+  <description></description>
+  <keepDependencies>false</keepDependencies>
+  <properties>
+    <com.coravy.hudson.plugins.github.GithubProjectProperty plugin="github@1.29.3">
+      <projectUrl>https://github.com/bartoszbetka/concent-deployment.git/</projectUrl>
+      <displayName></displayName>
+    </com.coravy.hudson.plugins.github.GithubProjectProperty>
+    <org.jenkinsci.plugins.workflow.job.properties.PipelineTriggersJobProperty>
+      <triggers>
+        <com.cloudbees.jenkins.GitHubPushTrigger plugin="github@1.29.3">
```
cameel: Is `1.29.3` the version of the plugin? If so, we probably want to have it in our version file.

concent-builder/jenkins-configuration-files/config.xml
```diff
@@ -0,0 +1,84 @@
...
+node {
+    stage (&apos;Checkout repository&apos;) {
+        dir(&apos;/home/jenkins/concent-deployment/&apos;) {
+            git url:
+            &apos;https://github.com/bartoszbetka/concent-deployment.git&apos;,
+            branch: &apos;dev&apos;
+        }
+        ansiblePlaybook( 
+            inventory: &apos;/home/jenkins/inventory&apos;, 
+            playbook: &apos;/home/jenkins/concent-deployment/concent-builder/set-concent-version.yml&apos;
```
cameel: You should use `{{ jenkins_home_dir }}` here.

concent-builder/consts.yml
```diff
@@ -1,5 +1,8 @@
...
-data_dir:          /var/concent-data
-build_dir:         /home/{{ shared_user }}/build/{{ cluster }}
-local_secret_dir:  ../../concent-secrets
-deployment_values: ../../concent-deployment-values
+shared_user:                   builder
+data_dir:                      /var/concent-data
+build_dir:                     /home/{{ shared_user }}/build/{{ cluster }}
+local_secret_dir:              ../../concent-secrets
+deployment_values:             ../../concent-deployment-values
+deployment_values_repository:  https://github.com/golemfactory/concent-deployment-values.git
```
cameel: Why are you not taking this address from our file that lists repositories?

concent-builder/configure.yml
```diff
@@ -58,9 +75,34 @@
...
+            owner: jenkins
+            group: jenkins
+
+        - name:   Add jenkins role to sudoers
+          copy:
+            src:  "jenkins-configuration-files/jenkins_sudoers"
+            dest: "/etc/sudoers.d/jenkins"
+
+        - name:   Add inventory file to run ansible playbook by jenkins
+          copy:
```
cameel: I think you should use `ansible_inventory` and just add `--connection=local` to the commandline where needed.

concent-builder/configure.yml
```diff
@@ -58,9 +75,34 @@
             state:  present
             groups: docker
 
+        - name:    Add jenkins user to sudo group
+          user:
```
cameel: I don't think giving jenkins sudo powers is a good idea.

concent-builder/configure-and-run-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+    - consts.yml
+    - repositories.yml
+    - ../containers/versions.yml
+    - "{{ deployment_values }}/var.yml"
+    - "{{ deployment_values }}/var-{{ cluster }}.yml"
+  tasks:
+    - become:      yes
+      become_user: jenkins
+      block:
+        - name: Copy concent and virtualmachine for deployment build
```
cameel: `virtualmachine` -> `virtual machine`

concent-builder/configure-and-run-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+          command: ./wait-until-ready.sh geth 600
+          args:
+            chdir:  "{{ build_dir }}/concent-deployment/kubernetes/build/"
+
+    - become:      yes
+      become_user: jenkins
+      block:
+        - name:  Run end-to-end test
+          shell: source ../virtualenv/bin/activate &&                  \
+              ./api-e2e-tests.sh http://devel.concent.golem.network
```
cameel: The address of the cluster should not be hard-coded here.

concent-builder/configure-and-run-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+            group:      jenkins
+            remote_src: yes
+
+    - become:      yes
+      become_user: "{{ shared_user }}"
+      block:
+        - name: Wait for geth synchronization
+          command: ./wait-until-ready.sh geth 600
+          args:
+            chdir:  "{{ build_dir }}/concent-deployment/kubernetes/build/"
```
cameel: Waiting for geth to get synchronized should be a separate step in the pipeline. We want to be able to see that it's the synchronization that failed and not the tests themselves.

concent-builder/configure-and-run-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+    - "{{ deployment_values }}/var.yml"
+    - "{{ deployment_values }}/var-{{ cluster }}.yml"
+  tasks:
+    - become:      yes
+      become_user: jenkins
+      block:
+        - name: Copy concent and virtualmachine for deployment build
+          command: cp                                                                  \
+              --recursive                                                              \
+              "{{ build_dir }}/concent-deployment/containers/build/{{ item.path }}"    \
```
cameel: Isn't there some kind of setup step in the pipeline where we could put this? I think this setup should not be a part of the step that runs the tests.

concent-builder/configure-and-run-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+            src:        "{{ jenkins_home_dir }}/concent-deployment/concent-builder/jenkins-configuration-files/local_settings.py"
+            dest:       "{{ jenkins_home_dir }}/concent/concent_api/concent_api/settings/local_settings.py"
+            owner:      jenkins
+            group:      jenkins
+            remote_src: yes
+
+    - become:      yes
+      become_user: "{{ shared_user }}"
+      block:
+        - name: Wait for geth synchronization
```
cameel: Jenkins should not have to run anything as a different user. This would require sudo access.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+        - name: Clone concent repositories
+          git:
+            repo:    "{{ item.url }}"
+            dest:    "{{ jenkins_home_dir }}/{{ item.name }}"
+            clone:   yes
+            update:  yes
+            version: dev
+            force:   yes
+          with_items:
+          #  - { url: "{{ repositories['concent-deployment'].url }}", name: concent-deployment }
```
cameel: Why is it commented-out?

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            - { url: "https://github.com/bartoszbetka/concent-deployment.git", name: concent-deployment }
```
cameel: The URL should not be hard-coded.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            - { url: "{{ deployment_values_repository }}", name: concent-deployment-values }
+
+        - name: Create directory for concent jenkins secrets
+          file:
+            path:  "{{ jenkins_home_dir }}/{{ item }}"
+            state: directory
+            owner: jenkins
+            group: jenkins
```
cameel: You're running in a block with `become_user: jenkins`. This is not necessary.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            - concent-secrets/cloud/
+            - concent-secrets/jenkins/
+
+    - become:      yes
+      become_user: root
+      block:
+        - name: Upload secrets
+          copy:
+            src:   "{{ local_secret_dir }}/{{ cluster }}/{{ item }}"
+            dest:  "{{ jenkins_home_dir }}/concent-secrets/{{ cluster }}/{{ item }}"
```
cameel: Do we really need to copy the secrets to Jenkins directory? Can't it use the ones from `/var/`?

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            - "/var/lib/jenkins/jobs/Build-deploy-and-test/workspace"
+            - "/var/lib/jenkins/users/admin/"
+            - "/var/lib/jenkins/secrets/"
+
+        - name: Upload jenkins secrets
+          copy:
+            src:   "{{ local_secret_dir }}/jenkins/{{ item.name }}"
+            dest:  "/var/lib/jenkins{{ item.path }}{{ item.name }}"
+            owner: "jenkins"
+            group: "jenkins"
```
cameel: Please make the secrets readable only for the `jenkins` user.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+          with_items:
+            - { path: "/users/admin/", name: "config.xml" }
+            - { path: "/users/admin/", name: "apiTokenStats.xml" }
+            - { path: "/", name: "credentials.xml" }
+            - { path: "/secrets/", name: "master.key" }
+            - { path: "/secrets/", name: "hudson.util.Secret" }
+
+        - name: Upload jenkins configuration file
+          # FIXME This need use as remote_src
```
cameel: What about this `FIXME`?

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            group: "jenkins"
+          with_items:
+            - { path: "/jobs/Build-deploy-and-test/", name: "config.xml" }
+            - { path: "/", name: "github-plugin-configuration.xml" }
+            - { path: "/", name: "jenkins.model.JenkinsLocationConfiguration.xml" }
+
+        - name: Restart jenkins service
+          service:
+            enabled: yes
+            name:    jenkins
```
cameel: You're not just restarting it. You're also enabling it. Please include that information in task name.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            state:   restarted
+
+    - become:      yes
+      become_user: jenkins
+      block:
+        - name: Wait for jenkins after restart daemon
```
cameel: `restart daemon` -> `restarting the daemon`

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+          wait_for:
+            port:  8080
+            delay: 60
+
+        - name: Download jenkins-cli
+          get_url:
+            url:   "http://{{ jenkins_address }}/jnlpJars/jenkins-cli.jar"
+            dest:  "{{ jenkins_home_dir }}/jenkins-cli.jar"
```
cameel: Please include Jenkins version in the file name. Otherwise this playbook won't download a newer version when we update Jenkins.

BTW, is the file stored on our server or is it just a redirect to another machine? If it's stored locally you could just get a path to it and run it directly without downloading it. This would solve the problem of updating it - it would get updated automatically with Jenkins.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            mode:  0755
+
+        - name: Install needed plugin for jenkins
+          command: java                                               \
+              -jar jenkins-cli.jar                                    \
+              -s http://localhost:8080/                               \
+              -auth @/home/jenkins/concent-secrets/jenkins/.token     \
```
cameel: What is this token for?

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+              install-plugin                                          \
+              "{{ item }}"
+          args:
+            chdir: "{{ jenkins_home_dir }}/"
+          with_items:
+            - workflow-aggregator
+            - ansible
+
+        - name: Install github plugin for jenkins
+          command: java                                               \
```
cameel: This is this plugin being installed separately from the two listed above?

cameel: Does this reinstall the plugin if it's already installed? Does it update it if a new version is released?

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+              github-oauth
+          args:
+            chdir: "{{ jenkins_home_dir }}/"
+
+    - become:      yes
+      become_user: root
+      block:
+        - name: Restart jenkins service
+          service:
+            enabled: yes
```
cameel: Why are you enabling it again? This only need to be done once. Actually, it whould be better to have a separate step for enabling it.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+            name:    jenkins
+            state:   restarted
+
+        - name: Add nginx config
+          copy:
+            src:   "jenkins-configuration-files/default"
+            dest:  "/etc/nginx/sites-enabled/default"
```
cameel: I'd rather not put the configuration in `default`. Please call it `jenkins`.

You may also need to remove `default` but you should do this in the playbook that installs `nginx`. I.e. in `configure.yml`.

cameel: You should put the configuration in `sites-available` and create a link to it in `sites-enabled`.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+        - name: Restart nginx service
+          service:
+            enabled: yes
+            name:    nginx
+            state:   restarted
+
+    - become:      yes
+      become_user: "{{ shared_user }}"
+      block:
+        - name: Check if firewall rule that allow http traffic exist
```
cameel: `allow http traffic exist` -> `allows HTTP traffic exists`

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,184 @@
...
+        - name: Add firewall rule to allow http traffic
+          command: gcloud compute firewall-rules create \
+              allow-http-traffic                        \
+              --allow=tcp:80                            \
+              --priority=1000                           \
+              --direction=ingress                       \
+              --target-tags=http-server
+          when: result.stdout == ""
+
+        - name: Add firewall rule to deployment server
```
cameel: `Add ... to` -> `Enable ... for`

concent-builder/jenkins-configuration-files/config.xml
```diff
@@ -0,0 +1,84 @@
+<?xml version='1.1' encoding='UTF-8'?>
```
cameel: Please move `concent-builder/jenkins-configuration-files/` to `concent-builder/templates/jenkins/` or `concent-builder/files/jenkins/` (depending whether you use templates or static files).

concent-builder/jenkins-configuration-files/github-plugin-configuration.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<github-plugin-configuration plugin="github@1.29.3">
+  <configs/>
+  <hookUrl>http://builder.concent.golem.network/github-webhook/</hookUrl>
+  <hookSecretConfig>
+    <credentialsId>f3d86e33-7dbe-43a8-bacd-d131120e998b</credentialsId>
+  </hookSecretConfig>
```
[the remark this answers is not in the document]
bartoszbetka: This is secret ID that must be synchronized with value in the secret file. We don't have template system for secrets file so I think the better option is to stay hard coded this value as in the secret file.

concent-builder/jenkins-configuration-files/config.xml
```diff
@@ -0,0 +1,84 @@
...
+            inventory: &apos;/home/jenkins/inventory&apos;, 
+            playbook: &apos;/home/jenkins/concent-deployment/concent-builder/configure-and-run-tests.yml&apos;
+        )
+    }
+}</script>
+    <sandbox>true</sandbox>
+  </definition>
+  <triggers/>
+  <disabled>false</disabled>
+</flow-definition>
```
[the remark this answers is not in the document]
bartoszbetka: Why you want 4-character here ? This is default configuration file from jenkins.

concent-builder/jenkins-configuration-files/config.xml
bartoszbetka: This information is included in the pull request description.

concent-builder/consts.yml
bartoszbetka: This is value of `concent-deployment-values` repository which is not included in our lists of repository.

concent-builder/jenkins-configuration-files/github-plugin-configuration.xml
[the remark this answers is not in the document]
cameel: No way :) Secrets must never be hard-coded in the repo. If our deployment system does not allow for that, we change our system.

The question is - what do we have to change? We have secrets in .yml files so you should be able to load them in Ansible and use them when rendering a template...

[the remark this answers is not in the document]
cameel: Oh, but by secretID you mean something like the name of the secret? It's fine then, though I'd prefer a comment explaining this here.

concent-builder/jenkins-configuration-files/config.xml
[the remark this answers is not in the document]
cameel: Because we want this file to be edited by us with a text editor (not automatically regenerated using Jenkins GUI) and we use 4-space indents everywhere (except formats like YAML where syntax rules do not let us).

It's fine to keep the default config in a separate commit but please then refactor it in another commit.

concent-builder/jenkins-configuration-files/config.xml
cameel: OK. But I'll leave it unresolved until then just as a reminder.

concent-builder/configure.yml
bartoszbetka: Ok, my mistake here. However, we need add jenkins to sudoers to specify no passwd when the playbooks run tasks as root user. Maybe we need create different user for running playbooks ?

concent-builder/configure.yml
bartoszbetka: Ok. I found the way to do this by ansible plugin.

cameel: Why do you need a plugin for this? It works for me without any plugins.

concent-builder/configure.yml
cameel: The playbook that installs and configures jenkins is expected to be run by a user (e.g. your or my user) who has sudo access. But this needs to be done only once.

The playbook that is started by jenkins and runs tests should not have or need sudo access at all. If it does (e.g. because it tries to access files of a different user (e.g. `builder`), we need to change this). For example the pull request that allows building from accounts other than `builder` can help with that.

concent-builder/configure-and-run-tests.yml
bartoszbetka: I corrected the name, this is virtual environment.

concent-builder/configure-and-run-tests.yml
bartoszbetka: Do we still want to wait for geth? After we merged pull request with liveness and readiness probe for geth, the deployment process wait 5 minutes for geth synchronous and return timeout when time exceed.

cameel: In that case you can ignore it here. We'll come back to this in the future if it turns out that a separate waiting step would be more convenient.

concent-builder/configure-and-run-tests.yml
bartoszbetka: This task depends on the `build-test-and-push` playbook, so we can add it there. However I think that it's in the right place. The playbook creates independent environment that copies a specific concent version with a specific version of the tests, launches them and then deletes them.

concent-builder/configure-and-run-tests.yml
bartoszbetka: I think that it's not possible to do, maybe after this pull request #223 .

concent-builder/configure-and-run-tests.yml
cameel: The problem is that if this initialization fails, it's counted as a failure of the code being tested. I'd prefer to have separate initialize/test/cleanup steps. May be separate playbooks if there's no other way.

concent-builder/configure-jenkins.yml
bartoszbetka: I used forked repository for github webhook tests. Please read description of pull request.

concent-builder/configure-and-run-tests.yml
cameel: Then we have to do #223 first. I has already been reviewed by me and is waiting for corrections.

concent-builder/configure-jenkins.yml
bartoszbetka: As above. The commented version is correct, the forked repository is for testing purpose.

concent-builder/configure-jenkins.yml
cameel: It's not that obvious that this is specifically the part the description is talking about. And anyway, there's more than one such place in the code. It would be better if you marked things like this with something like `# TMP:` in the code with a comment that explains it. It would make those places easier to find and would also make the purpose clear in the review.

concent-builder/configure-jenkins.yml
bartoszbetka: This is question for separate issue and changes deployment process.

concent-builder/configure-jenkins.yml
bartoszbetka: It is enabled by default after installation, so there is not change here.

concent-builder/configure-jenkins.yml
bartoszbetka: Done. The task with template module doesn't have to `option remote_src`, so it gets it from the local source.

concent-builder/configure-jenkins.yml
bartoszbetka: https://wiki.jenkins.io/display/JENKINS/Jenkins+CLI#JenkinsCLI-ObtainingtheCLIclient

concent-builder/configure-jenkins.yml
bartoszbetka: The token allow to authentication to the jenkins-cli which is allow to changing jenkins configuration. I think that jenkins-cli communicate to jenkins by api.

concent-builder/configure-jenkins.yml
cameel: Why though? You're copying them here. It's not a part of a different playbook.

concent-builder/configure-jenkins.yml
cameel: If you're not changing it, it would be clearer to just not include this option at all.

concent-builder/configure-jenkins.yml
cameel: Well, OK. I wanted to suggest to disable remote access and simply have no password (or a dummy password) when connecting locally but I realized that it would allow any user on the system (not just `jenkins` or `root`) to connect to it. So maybe it's better to have a token after all.

cameel: One more thing.

You're running Jenkins on `localhost` rather than `0.0.0.0` which is good because the system won't allow remote connections. But just in case it would be a good idea to disable remote access in configuration too. See:
- https://stackoverflow.com/questions/44501596/jenkins-disable-cli-over-remoting-via-a-groovy-script/45250763#45250763
- https://support.cloudbees.com/hc/en-us/articles/234709648-Disable-Jenkins-CLI

concent-builder/configure-jenkins.yml
cameel: It does not really say if it's storing it on the sever. I only see that you're using a local URL but the server could just as well be redirecting you to some remote location.

Could you try to find it on disk when Jenkins is installed?

concent-builder/configure-jenkins.yml
bartoszbetka: Why do you want to complicate it in this way ?

bartoszbetka: I think that if we want change the current solution, we could add `jenkins.conf` file to the `/etc/nginx/conf.d/` directory and delete the `/etc/nginx/sites-enabled/default` config.

cameel: > Why do you want to complicate it in this way ?

Because that's the standard way to do it. If you use things the way they're supposed to be used, it's always easier for other people to understand your stuff. Someone familiar with how nginx works would naturally assume that `sites-enabled` contains just links and could remove the configuration file by mistake thinking it's just a link.

I think it fits more in `sites-available`/`sites-enabled`. That's where you're supposed to put your `server` blocks. `conf.d` sounds like something where you would put bits of global configuration that pertain to all the sites

concent-builder/jenkins-configuration-files/config.xml
bartoszbetka: This problem will be resolve in separate issue #287.

concent-builder/configure-jenkins.yml
bartoszbetka: The solution from your comment is not that obvious. Do we want assume that `/var/concent-data` path is always exist ? That may work, but we must run the `cluster-deploy-secrets.yml` playbook in specific case.  I think that the `configure-jenkins.yml` playbook should be independent.

concent-builder/configure-jenkins.yml
bartoszbetka: I found the `jenkins-cli.jar` file on the local disk. The task now copy this file to home jenkins directory.

concent-builder/configure-jenkins.yml
bartoszbetka: I check the time of files and they are not change, so it's look like the jenkins not reinstall plugins that are already installed. When the plugin is updated then new one version is installed.

concent-builder/configure-jenkins.yml
bartoszbetka: This is a default setting in my case https://stackoverflow.com/questions/44501596/jenkins-disable-cli-over-remoting-via-a-groovy-script/45250763#45250763

concent-builder/configure-and-run-tests.yml
bartoszbetka: The `cleanup` step must depends on `test`step. When the `test` playbook fails, the `cleanup` playbook will not be executed. I maybe find solution how to execute always playbook in jenkins but I think we should keep this in the one playbook.

concent-builder/configure-jenkins.yml
cameel: So you mean that it is disabled by default? How do you know that? That post on SO only says how to disable it, does not say anything about defaults.

concent-builder/files/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,76 @@
+properties([[$class:
+'GithubProjectProperty',
+    displayName: '',
+    projectUrlStr: 'https://github.com/bartoszbetka/concent-deployment.git/'],
```
cameel: This is wrapped in a weird way. Hard to see what is nested in what. Why not do this?
``` groovy
properties([
    [
        $class:        'GithubProjectProperty',
        displayName:   '',
        projectUrlStr: 'https://github.com/bartoszbetka/concent-deployment.git/'
    ],
    pipelineTriggers([githubPush()])
])
```

concent-builder/files/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,76 @@
...
+pipelineTriggers([githubPush()])])
+
+stage ('Checkout repository') {
+    dir('/home/jenkins/concent-deployment/') {
+        git url:
+        'https://github.com/bartoszbetka/concent-deployment.git',
```
cameel: We already have the CI hook enabled in the repo at golemfactory, so I think this should now point at that repo rather than your fork.

concent-builder/files/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,76 @@
...
+    ansiblePlaybook(
+        extraVars: [
+            cluster: 'concent-dev',
+        ],
+        inventory: '/home/jenkins/concent-deployment-values/ansible_inventory',
+        playbook:  '/home/jenkins/concent-deployment/concent-builder/deploy.yml',
+        extras:    '--connection local'
+    )
+}
+stage ('Initialized Test Environment') {
```
cameel: `Initialized` -> `Initialize`

concent-builder/files/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,76 @@
...
+    ansiblePlaybook(
+        extraVars: [
+            cluster: 'concent-dev',
+        ],
+        inventory: '/home/jenkins/concent-deployment-values/ansible_inventory',
+        playbook:  '/home/jenkins/concent-deployment/concent-builder/initialized-test-environment.yml',
```
cameel: `initialized-test-environment.yml` -> `initialize-test-environment.yml`

concent-builder/files/jenkins/github-plugin-configuration.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<github-plugin-configuration plugin="github@1.29.3">
+  <configs/>
+  <hookUrl>http://builder.concent.golem.network/github-webhook/</hookUrl>
```
cameel: The domain (or even the whole URL) should come from concent-deployment-values and not be hard-coded.

concent-builder/files/jenkins/github-plugin-configuration.xml
```diff
@@ -0,0 +1,9 @@
...
+  <hookSecretConfig>
+    <!-- This is the identifier of the specific secret-->
+    <credentialsId>bbe1e591-1dda-4398-aa6c-1a8a5b4ca634</credentialsId>
```
cameel: Could you mention also the name of the file that should contain the secret?

concent-builder/configure-jenkins.yml
cameel: It's still not fixed. The file is still in `sites-enabled`.

concent-builder/configure-jenkins.yml
cameel: Please make it a symlink.

concent-builder/configure-jenkins.yml
cameel: What about this? You're still not setting `mode` here.

concent-builder/configure-jenkins.yml
cameel: It's fine to assume that `/var/concent-data/`, but you're right - I don't want to require running `cluster-deploy-secrets.yml`. We need a separate playbook. But in that case let's leave it as is for now. We can add that playbook later.

concent-builder/configure-jenkins.yml
```diff
@@ -0,0 +1,170 @@
...
+          copy:
+            src:   "{{ local_secret_dir }}/jenkins/{{ item.name }}"
+            dest:  "/var/lib/jenkins{{ item.path }}{{ item.name }}"
+            owner: "jenkins"
+            group: "jenkins"
+          with_items:
+            - { path: "/users/admin/", name: "config.xml" }
+            - { path: "/users/admin/", name: "apiTokenStats.xml" }
+            - { path: "/", name: "credentials.xml" }
+            - { path: "/secrets/", name: "master.key" }
```
cameel: Right now these look like absolute paths which is confusing.

Please add the trailing slash in `dest`:
``` yaml
            dest:  "/var/lib/jenkins{{ item.path }}{{ item.name }}/"
```
and use relative paths here.

cameel: I think it would be better to have same subdirectories in our secret dir. E.g. store `config.xml` in `secrets/jenkins/users/admin/` rather than in `secrets/jenkins/`.

Then you would be able to simplify this to:
``` yaml
            ...
            src:   "{{ local_secret_dir }}/jenkins/{{ item }}"
            dest:  "/var/lib/jenkins/{{ item }}"
            ...
          with_items:
            - users/admin/config.xml
            - users/admin/apiTokenStats.xml
            - credentials.xml
           ...
```

concent-builder/configure-jenkins.yml
cameel: We have the hook. This should be changed now.

concent-builder/files/jenkins/Jenkinsfile
cameel: We have the hook. This should be changed to the main repository URL now.

concent-builder/files/jenkins/local_settings.py
```diff
@@ -0,0 +1,3 @@
+from .development import *
+
+STORAGE_CLUSTER_ADDRESS = 'http://devel.concent.golem.network/'
```
cameel: Plase make it a template and take the address from concent-deployment-values.

concent-builder/files/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,76 @@
...
+
+stage ('Checkout repository') {
+    dir('/home/jenkins/concent-deployment/') {
+        git url:
+        'https://github.com/bartoszbetka/concent-deployment.git',
+        branch: 'CI'
+    }
+    ansiblePlaybook(
+        inventory: '/home/jenkins/concent-deployment-values/ansible_inventory',
+        playbook:  '/home/jenkins/concent-deployment/concent-builder/set-concent-version.yml',
```
cameel: Please make this file a template so that we don't have to hard-code `/home/jenkins/`

concent-builder/template/jenkins/jenkins.model.JenkinsLocationConfiguration.xml.j2
```diff
@@ -0,0 +1,5 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<jenkins.model.JenkinsLocationConfiguration>
+  <adminAddress>{{ jenkins_admin_email_address }}</adminAddress>
+  <jenkinsUrl>http://{{ jenkins_url_address }}/</jenkinsUrl>
```
cameel: `jenkins_url_address` -> `jenkins_url`. And I'd include the `http://` part in the variable too.

concent-builder/files/jenkins/Jenkinsfile
cameel: The URL should come from our YAML config that contains repository addresses.

concent-builder/configure-and-run-tests.yml
cameel: Well, OK let's keep it like this for now but I'd like to find a solution to make it more modular in the future.

concent-builder/configure-and-run-tests.yml
cameel: I this done? I don't see this in the code any more.

concent-builder/configure.yml
```diff
@@ -64,3 +81,32 @@
...
+          file:
+            path:  "{{ jenkins_home_dir }}/"
+            state: directory
+            owner: jenkins
+            group: jenkins
+
+        - name:   Add jenkins role to sudoers
+          copy:
+            src:  "files/jenkins/jenkins_sudoers"
+            dest: "/etc/sudoers.d/jenkins"
```
cameel: This is no longer necessary. I see that you have changed the code so that jenkins no longer needs to do anything as root.

concent-builder/configure.yml
cameel: Is it fixed?

https://github.com/golemfactory/concent-deployment/pull/284#discussion_r246439542

concent-builder/configure-jenkins.yml
bartoszbetka: I don't see it.

concent-builder/configure-jenkins.yml
bartoszbetka: Because I checked what is in this file after jenkins installation.

concent-builder/configure.yml
bartoszbetka: The problem is in the `install-repository.yml` playbook when we clone repositories to `/var/concent-data`, we still use the `builder` user.

concent-builder/configure-and-run-tests.yml
bartoszbetka: This is not part of code any more.

concent-builder/files/jenkins/Jenkinsfile
bartoszbetka: This change is in a CI branch, I keep old fork for tests

concent-builder/configure-jenkins.yml
bartoszbetka: This is absolute paths. Why you want use trailing slash after name of file ?

bartoszbetka: Ok, I think you want get this ?

```yaml
          copy:
            src:   "{{ local_secret_dir }}/jenkins/{{ item.name }}"
            dest:  "/var/lib/jenkins/{{ item.path }}{{ item.name }}"
            owner: "jenkins"
            group: "jenkins"
          with_items:
            - { path: "users/admin/", name: "config.xml" }
            - { path: "users/admin/", name: "apiTokenStats.xml" }
            - { path: "", name: "credentials.xml" }
            - { path: "secrets/", name: "master.key" }
            - { path: "secrets/", name: "hudson.util.Secret" }

```

concent-builder/files/jenkins/Jenkinsfile
bartoszbetka: I don't remember exactly but I think in this file there is a problem with: `{` and `}` characters.

concent-builder/files/jenkins/Jenkinsfile
bartoszbetka: The same https://github.com/golemfactory/concent-deployment/pull/284#discussion_r249460766

concent-builder/files/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,76 @@
...
+        extras:    '--connection local'
+   )
+}
+stage ('Build') {
+    ansiblePlaybook(
+        extraVars: [
+            cluster: 'concent-dev',
+        ],
+        inventory: '/home/jenkins/concent-deployment-values/ansible_inventory',
+        playbook:  '/home/jenkins/concent-deployment/concent-builder/install-repositories.yml',
```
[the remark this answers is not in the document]
bartoszbetka: Jenkins github plugin do fetch and checkout. The `install-repositories.yml` only updates repositories that we use in `build-test-and-push.yml` playbook. After the pull request #223 was merged, we don't need checkout `concent-deployment repository` in `build-test-and-push.yml` playbook.

concent-builder/initialize-test-environment.yml
```diff
@@ -0,0 +1,27 @@
...
+          with_items:
+            - { path: "repositories/concent", name: "concent" }
+            - { path: "virtualenv", name: "virtualenv" }
+
+        - name: Copy local_settings.py file
+          template:
+            src:        "template/jenkins/local_settings.py.j2"
+            dest:       "{{ jenkins_home_dir }}/concent/concent_api/concent_api/settings/local_settings.py"
+            owner:      jenkins
+            group:      jenkins
```
cameel: This is not necessary. You're in a block with `become_user: jenkins`.

concent-builder/configure-jenkins.yml
cameel: It's gone now. Maybe you have fixed it in the meantime.

But I've found another one: https://github.com/golemfactory/concent-deployment/pull/284#pullrequestreview-194706063

concent-builder/configure.yml
cameel: #191 was meant to fix that. In that case you can leave it as is in this pull request but please create an issue to take away Jenkins' sudo access as soon as #191 is resolved.

concent-builder/files/jenkins/Jenkinsfile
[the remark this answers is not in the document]
cameel: OK. Please add a comment explaining that.

concent-builder/files/jenkins/config.xml
```diff
@@ -0,0 +1,41 @@
...
+  </views>
+  <primaryView>all</primaryView>
+  <slaveAgentPort>-1</slaveAgentPort>
+  <label></label>
+  <crumbIssuer class="hudson.security.csrf.DefaultCrumbIssuer">
+    <excludeClientIPFromCrumb>false</excludeClientIPFromCrumb>
+  </crumbIssuer>
+  <nodeProperties/>
+  <globalNodeProperties/>
+</hudson>
```
cameel: This file does not follow our coding style. Indents should be 4-spaces long. Please also align stuff for readabilty. These files are meant to be edited by hand so they should be cleaner.

bartoszbetka: This is a new file that was added after we decided to disable logging into Jenkins. Done.

cameel: Sure. But it does not make it exempt from our rules :)

PR #290, opened 2019-01-07
Add option that allow enabling routing based on the golem-messages version with a single cluster

### LINKED ISSUE #289: Allow enabling routing based on golem-messages version even with a single cluster
Our nginx-proxy can be configured to route HTTP messages based on golem-messages version included in the `X-Golem-Messages` (#258).

Currently there are only two possibilities: either there are exactly two GM versions available or the feature is completely disabled. The problem is that disabling it changes behavior of Concent endpoints because it disables all the checks against the header. When it's enabled, messages with GM version that's not available result in a HTTP 404. That's not the case when it's disabled. This makes the behavior inconsistent.

To alleviate this, we need to allow enabling routing even when there's only one cluster with one GM version available.

bartoszbetka:
Resolves #289

### DIFF
--- a/kubernetes/config-maps/nginx-proxy/configs/http.conf.j2
+++ b/kubernetes/config-maps/nginx-proxy/configs/http.conf.j2
@@ -20,15 +20,21 @@ upstream {{ current_golem_messages_version }} {
      server nginx-proxy.default.svc.cluster.local;
 }
 
+# The previous_golem_messages_version variable controls routing of the previous golem-messages version to the second cluster.
+# When the variable is empty, the traffic to second cluster is disabled and only the current golem-messages version is supported.
+{% if previous_golem_messages_version is not none %}
 upstream {{ previous_golem_messages_version }} {
      server {{ concent_versions[previous_concent_version].cluster_address }}:443;
 }
+{% endif %}
 
 # Map to different upstream backends based on the `X-Golem-Messages` http header
 map $http_x_golem_messages $version {
      ""          "http://{{ current_golem_messages_version }}";
      "~^{{ current_golem_messages_version }}.*$"   "http://{{ current_golem_messages_version }}";
+     {% if previous_golem_messages_version is not none %}
      "~^{{ previous_golem_messages_version }}.*$"  "https://{{ previous_golem_messages_version }}";
+     {% endif %}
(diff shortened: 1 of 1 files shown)


kubernetes/config-maps/nginx-proxy/configs/http.conf.j2
```diff
@@ -20,15 +20,19 @@ upstream {{ current_golem_messages_version }} {
...
 }
+{% endif %}
 
 # Map to different upstream backends based on the `X-Golem-Messages` http header
 map $http_x_golem_messages $version {
      ""          "http://{{ current_golem_messages_version }}";
      "~^{{ current_golem_messages_version }}.*$"   "http://{{ current_golem_messages_version }}";
+     {% if previous_golem_messages_version is not none and previous_golem_messages_version is defined %}
      "~^{{ previous_golem_messages_version }}.*$"  "https://{{ previous_golem_messages_version }}";
+     {% endif %}
```
cameel: This part should have a comment explaining why one connection is over HTTPS and the other over HTTP.

kubernetes/config-maps/nginx-proxy/configs/http.conf.j2
```diff
@@ -20,15 +20,19 @@ upstream {{ current_golem_messages_version }} {
      server nginx-proxy.default.svc.cluster.local;
 }
 
+{% if previous_golem_messages_version is not none and previous_golem_messages_version is defined %}
 upstream {{ previous_golem_messages_version }} {
      server {{ concent_versions[previous_concent_version].cluster_address }}:443;
 }
```
cameel: This should have a comment explaining why this is enabled only when `previous_golem_messages_version` is non-empty.

kubernetes/config-maps/nginx-proxy/configs/http.conf.j2
```diff
@@ -20,15 +20,19 @@ upstream {{ current_golem_messages_version }} {
      server nginx-proxy.default.svc.cluster.local;
 }
 
+{% if previous_golem_messages_version is not none and previous_golem_messages_version is defined %}
```
cameel: You should assume that every config variable is always defined. It may be empty but not undefined.

PR #298, opened 2019-01-09
Downgrade pip version to 18.0, newest version of pip on debian jessie causes `403 Client Error` error when `--extra-index-url` flag is used. 

### LINKED ISSUE #296: pip returns `403 Client Error: SSL is required` error when building verifier image.
We use docker with blender image from dockerhub golemfactory/blender. The some changes in debian jessie libs cause error when we try install [requirements](https://github.com/golemfactory/golem/blob/b0.19/apps/rendering/resources/images/requirements.txt) for verifier. 

To reproduce problem:
```
python3 -m pip install --no-cache-dir --extra-index-url https://builds.golem.network opencv-python
Looking in indexes: https://pypi.org/simple, https://builds.golem.network
Collecting opencv-python
Could not install packages due to an EnvironmentError: 403 Client Error: SSL is required for url: http://pypi.python.org/simple/opencv-python/

### PR DESCRIPTION
**Pull request by bartoszbetka**
The docker verifier image that used debian jessie needs pip version 18.0 or older.The new version of pip cause `403 Client Error: SSL is required for url: http://pypi.python.org/simple/opencv-python/` error, when the `--extra-index-url` flag is used with HTTPS protocol and the package is from base python source.

Resolves #296

### DIFF
--- a/containers/versions.yml
+++ b/containers/versions.yml
@@ -18,3 +18,7 @@ kubectl_version:              "1.13.0-00"
 python_version:               "3.6"
 rabbitmq_alpine_version:      "3.7.6"
 su_exec_hash:                 "60e8c3010aaa85f5d919448d082ecdf6e8b75a1c"
+# The docker verifier image that uses debian jessie needs pip version 18.0 or older.
+# Newer versions of pip fail with `403 Client Error: SSL is required for url: http://pypi.python.org/simple/opencv-python/` error,
+# when the `--extra-index-url` flag is used with HTTPS protocol and the package is from base python source.
+verifier_image_pip_version:   "18.0"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: containers/versions.yml [hunk 1]
```diff
@@ -18,3 +18,7 @@ kubectl_version:              "1.13.0-00"
 python_version:               "3.6"
 rabbitmq_alpine_version:      "3.7.6"
 su_exec_hash:                 "60e8c3010aaa85f5d919448d082ecdf6e8b75a1c"
+# The docker verifier image that used debian jessie needs pip version 18.0 or older.
+# The new version of pip cause `403 Client Error: SSL is required for url: http://pypi.python.org/simple/opencv-python/` error,
+# when the `--extra-index-url` flag is used with HTTPS protocol and the package is from base python source.
```
cameel: `used` -> `uses`
`The new version of pip cause` -> `Newer versions of pip fail with`

PR #307, opened 2019-01-16
Change installing rustc from testing release

kbeker:
Change the way how rustc is installed because actual version which was installed `1.24.1` is not supported in golems `setup.py` and have to be updated to the newest release which is available only on testing release


cameel:
@kbeker You did not set the milestone.

kbeker:
@cameel Sorry, fixed :-)

PR #317, opened 2019-01-24
Separate installation of `middleman_protocol` as new layer in docker images

### LINKED ISSUE #291: Optimize docker images to reduce build times
Currently it takes 15 minutes to rebuild our containers in CI even with with Docker's caching. It's likely that we can reorder layers in the images to make caching more effective.

Having properly working cache would also help in cases where a rebuild is not necessary without having to check if the image already exists. If you run `concent-builder/build-test-and-push.yml` again on the same version, the build should be instant.

bartoszbetka:
Resolves #291

### DIFF
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -27,27 +27,30 @@ GETH_SOURCE :=                              \
     build/geth/Dockerfile                   \
     build/geth/is-geth-synchronized.sh
 
-CONCENT_API_SOURCE :=                                \
-    build/concent-api/install-concent-api.sh         \
-    build/concent-api/install-dependencies.sh        \
-    build/concent-api/Dockerfile                     \
-    build/concent-api/configure-pyenv.sh             \
-    build/concent-api/clean-pyenv-dependencies.sh    \
-    build/concent-api/concent-api-manage.sh          \
-    build/concent-api/run-celery.sh                  \
-    build/concent-api/pyenv-build-dependencies.txt   \
-    build/concent-api/run-gunicorn.sh                \
+CONCENT_API_SOURCE :=                                              \
+    build/concent-api/install-concent-api-python-dependencies.sh   \
+    build/concent-api/install-middleman-protocol.sh                \
+    build/concent-api/install-dependencies.sh                      \
+    build/concent-api/Dockerfile                                   \
+    build/concent-api/configure-pyenv.sh                           \
+    build/concent-api/clean-pyenv-dependencies.sh                  \
+    build/concent-api/concent-api-manage.sh                        \
+    build/concent-api/run-celery.sh                                \
+    build/concent-api/pyenv-build-dependencies.txt                 \
+    build/concent-api/run-gunicorn.sh                              \
     build/concent-api/local_settings.py
 
-VERIFIER_SOURCE_FROM_CONCENT_API :=                 \
-    build/verifier/install-concent-api.sh           \
-    build/verifier/concent-api-manage.sh            \
-    build/verifier/run-celery.sh                    \
-    build/verifier/clean-pyenv-dependencies.sh      \
-    build/verifier/configure-pyenv.sh               \
-    build/verifier/testing.pref                     \
-    build/verifier/requirements.lock                \
-    build/verifier/concent-api.tar                  \
+VERIFIER_SOURCE_FROM_CONCENT_API :=                              \
+    build/verifier/install-concent-api-python-dependencies.sh    \
+    build/verifier/install-middleman-protocol.sh                 \
+    build/verifier/concent-api-manage.sh                         \
+    build/verifier/run-celery.sh                                 \
+    build/verifier/clean-pyenv-dependencies.sh                   \
+    build/verifier/configure-pyenv.sh                            \
+    build/verifier/testing.pref                                  \
+    build/verifier/requirements.lock                             \
+    build/verifier/cacheable-requirements.lock                   \
+    build/verifier/concent-api.tar                               \
     build/verifier/local_settings.py
 
 VERIFIER_SOURCE :=                                 \
@@ -65,12 +68,17 @@ SIGNING_SERVICE_SOURCE :=                                  \
     build/signing-service/Dockerfile                       \
     build/signing-service/install-dependencies.sh          \
     build/signing-service/install-signing-service.sh       \
+    build/signing-service/install-middleman-protocol.sh    \
     build/signing-service/clean-container.sh               \
     build/signing-service/run-signing-service.sh           \
     build/signing-service/create-user.sh                   \
     build/signing-service/build-dependencies.txt           \
     build/signing-service/testing.pref                     \
     $(SIGNING_SERVICE_SOURCE_FROM_CONCENT_API)
+
+CACHEABLE_REQUIREMENTS_FILES :=                                        \
+    build/concent-api/cacheable-requirements.lock                      \
+    build/signing-service/signing-service/cacheable-requirements.lock
 
 ifndef IMAGE_VERSION
     IMAGE_VERSION := $(shell git describe --always --abbrev=16)
@@ -97,7 +105,7 @@ postgresql: build/ $(POSTGRESQL_SOURCE)
 	docker build --tag $(IMAGE_PREFIX)postgresql:$(IMAGE_VERSION) build/postgresql/
 	docker tag $(IMAGE_PREFIX)postgresql:$(IMAGE_VERSION) $(IMAGE_PREFIX)postgresql:latest
 
-concent-api: build/ $(CONCENT_API_SOURCE) build/concent-api/requirements.lock build/concent-api/testing.pref build/repositories/concent build/concent-api/middleman_protocol/
+concent-api: build/ $(CONCENT_API_SOURCE) build/concent-api/requirements.lock build/concent-api/testing.pref build/repositories/concent build/concent-api/middleman_protocol/ build/concent-api/middleman_protocol/RELEASE-VERSION build/concent-api/cacheable-requirements.lock
 	docker build --tag $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) build/concent-api/
 	docker tag $(IMAGE_PREFIX)concent-api:$(IMAGE_VERSION) $(IMAGE_PREFIX)concent-api:latest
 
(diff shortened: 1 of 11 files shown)


containers/Makefile
```diff
@@ -203,8 +209,8 @@ build/signing-service/signing-service/RELEASE-VERSION: build/signing-service/sig
 build/signing-service/signing-service/: build/repositories/concent
 	cp --recursive "$</signing_service/." "$@"
 
-build/signing-service/middleman_protocol/: build/repositories/concent
-	cp --recursive "$</middleman_protocol/." "$@"
+# build/signing-service/middleman_protocol/: build/repositories/concent
+	# cp --recursive "$</middleman_protocol/." "$@"
```
cameel: Why is this commented out?

If it's not needed, remove it.
If it's a workaround, there should be a comment explaining what you're working around.

containers/Makefile
```diff
@@ -149,6 +152,9 @@ build/concent-api/requirements.lock: build/concent-api/concent-api.tar
 		--directory build/concent-api/           \
 		concent_api/requirements.lock
 
+build/%/cacheable-requirements.lock: build/%/requirements.lock
+	awk '$$2!="../middleman_protocol"' "$<" > "$@"
```
cameel: Why `%` in the rule? This rule is specific to middleman. We don't want it to kick in if we add `cacheable-requirements.lock` elsewhere in the future.

containers/Makefile
```diff
@@ -99,7 +102,7 @@ verifier: build/ $(VERIFIER_SOURCE) build/repositories/golem/ build/verifier/gol
 	docker build --tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) build/verifier/
 	docker tag $(IMAGE_PREFIX)verifier:$(IMAGE_VERSION) $(IMAGE_PREFIX)verifier:latest
 
-signing-service: build/ $(SIGNING_SERVICE_SOURCE) build/signing-service/signing-service/ build/signing-service/middleman_protocol/ build/signing-service/signing-service/RELEASE-VERSION
+signing-service: build/ $(SIGNING_SERVICE_SOURCE) build/signing-service/signing-service/ build/signing-service/middleman_protocol/ build/signing-service/signing-service/RELEASE-VERSION build/signing-service/signing-service/cacheable-requirements.lock
 	docker build --tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) build/signing-service/
 	docker tag $(IMAGE_PREFIX)signing-service:$(IMAGE_VERSION) $(IMAGE_PREFIX)signing-service:latest
```
cameel: Shouldn't `dist/concent-signing-service-%.tar.xz` depend on `cacheable-requirements.lock` too?

containers/Makefile
```diff
@@ -194,7 +200,7 @@ build/repositories/concent: build/package-builder/clone-or-update.sh /var/concen
 build/repositories/golem/: build/package-builder/clone-or-update.sh /var/concent-data/golem
 	build/package-builder/clone-or-update.sh  "golem" "$@"
 
-build/%/middleman_protocol/: build/repositories/concent
+build/%/middleman_protocol/: build/repositories/concent build/virtualenv/
 	cp --recursive "$</middleman_protocol/." "$@"
```
cameel: Why does `middleman_protocol/` require having a virtualenv with Concent's requirements on the host?

You're only making a copy of middleman's source. You're not running it. This should not require installing any dependencies. If any other target that depends on `build/%/middleman_protocol/` needs this virtualenv then it should be the dependency of that target.

containers/verifier/Dockerfile.j2
```diff
@@ -7,10 +7,12 @@ COPY install-dependencies.sh                  /tmp/
 RUN  /tmp/install-dependencies.sh
 
 # Build virtualenv and concent apps
-COPY requirements.lock              /srv/http/concent_api/
-COPY middleman_protocol             /srv/http/middleman_protocol/
+COPY cacheable-requirements.lock    /srv/http/concent_api/
 COPY install-concent-api.sh         /tmp/
 RUN  /tmp/install-concent-api.sh
+COPY middleman_protocol                   /srv/http/middleman_protocol/
+COPY install-middleman-protocol.sh  /tmp/
```
cameel: This is misaligned.

containers/concent-api/Dockerfile.j2
```diff
@@ -4,15 +4,16 @@ COPY testing.pref                   /etc/apt/preferences.d/
...
 RUN  /tmp/install-concent-api.sh
-ADD concent-api.tar                 /srv/http/
-COPY concent-api-manage.sh          /usr/local/bin/
-COPY run-celery.sh                  /usr/local/bin/
-COPY run-gunicorn.sh                /usr/local/bin/
-
-COPY local_settings.py              /srv/http/concent_api/concent_api/settings/
+COPY middleman_protocol                   /srv/http/middleman_protocol/
+COPY install-middleman-protocol.sh        /tmp/
+RUN  /tmp/install-middleman-protocol.sh
```
cameel: I'd put `middleman_protocol/` above Concent. It's going to change much less. And Concent depends on it.

containers/signing-service/Dockerfile.j2
```diff
@@ -3,19 +3,19 @@ FROM debian:{{ debian_version }}
...
-COPY build-dependencies.txt             /tmp/
-COPY install-dependencies.sh            /tmp/
-COPY install-signing-service.sh         /tmp/
-COPY clean-container.sh                 /tmp/
-COPY middleman_protocol/                /usr/lib/signing_service/middleman_protocol/
-COPY signing-service/RELEASE-VERSION    /usr/lib/signing_service/middleman_protocol/RELEASE-VERSION
+COPY signing-service/cacheable-requirements.lock  /usr/lib/signing_service/signing-service/
+COPY testing.pref                                 /etc/apt/preferences.d/
+COPY build-dependencies.txt                       /tmp/
+COPY install-dependencies.sh                      /tmp/
```
cameel: Python requirements should be after system dependencies. Otherwise you'll be reinstalling system packages every time the Python dependencies change - which is completely unnecessary.

containers/signing-service/Dockerfile.j2
```diff
@@ -3,19 +3,19 @@ FROM debian:{{ debian_version }}
...
+COPY install-signing-service.sh                   /tmp/
+COPY clean-container.sh                           /tmp/
 RUN  /tmp/install-signing-service.sh
+COPY middleman_protocol                           /usr/lib/signing_service/middleman_protocol/
+COPY install-middleman-protocol.sh                /tmp/
+RUN  /tmp/install-middleman-protocol.sh
```
cameel: `middleman_protocol` should be before Signing Service. It's going to change much less often.

containers/verifier/Dockerfile.j2
```diff
@@ -7,10 +7,12 @@ COPY install-dependencies.sh                  /tmp/
...
+RUN  /tmp/install-middleman-protocol.sh
```
cameel: `middleman_protocol` should be installed before Concent.

containers/Makefile
bartoszbetka: Remove

containers/Makefile
bartoszbetka: This rule is uses as dependency in `concent-api` and `signing-service`. The `%` is uses to not create separate rules. We could change name of this requirement or hard code two the same rules with different path as you want ?
I think that it's not necessary. If we want to go this direction, we must investigate all `%` in our `Makefile`.

containers/Makefile
bartoszbetka: The virtualenv is uses here for unification purpose. This create middleman package already. The problem is because some containers are copy this directory before running virtualenv and some of it after. The solution is maybe create docker `nginx-proxy` image which is uses `virtualenv` after all others.

containers/signing-service/Dockerfile.j2
bartoszbetka: This optimization separate `middleman_protocol` because it's always change. Why do you think the `singing service` is changing more than `middleman_protocol` ?

containers/verifier/Dockerfile.j2
bartoszbetka: This optimization separate `middleman_protocol` because it's always change. Why do you want to install it before Concent ? It's not possible(`depends on Concent requirements`) and cause that concent always will be installing.

containers/concent-api/Dockerfile.j2
cameel: You have this:

``` dockerfile
COPY cacheable-requirements.lock          /srv/http/concent_api/
COPY install-concent-api.sh               /tmp/
RUN  /tmp/install-concent-api.sh
COPY middleman_protocol                   /srv/http/middleman_protocol/
COPY install-middleman-protocol.sh        /tmp/
RUN  /tmp/install-middleman-protocol.sh
```

I'm suggesting this:
``` dockerfile
COPY cacheable-requirements.lock          /srv/http/concent_api/
COPY middleman_protocol                   /srv/http/middleman_protocol/
COPY install-middleman-protocol.sh        /tmp/
RUN  /tmp/install-middleman-protocol.sh
COPY install-concent-api.sh               /tmp/
RUN  /tmp/install-concent-api.sh
```

containers/Makefile
cameel: Right, so it's used in two places. Still, it should be more limited.

There's no problem with other `%` in the `Makefile`. They are meant to be generic. But this one is not it should work only for `concent-api` and `signing-service`. But you don't have to duplicate the rule to achieve that. You can use a [Static Pattern Rule](https://www.gnu.org/software/make/manual/make.html#Static-Pattern):

``` makefile
CACHEABLE_REQUIREMENTS_FILES :=                   \
    build/concent-api/cacheable-requirements.lock \
    build/signing-service/cacheable-requirements.lock

$(CACHEABLE_REQUIREMENTS_FILES): build/%/cacheable-requirements.lock: build/%/requirements.lock
	awk '$$2!="../middleman_protocol"' "$<" > "$@"
```

You have to repeat target names but not the commands.

containers/Makefile
cameel: Let's talk about it tomorrow. I need more details.

cameel: Conclusions from our discussion:
- You do this because after the installation of Python packages in the virtualenv an `.egg-info` directory gets created inside `middleman_protocol/` and you want to have it in the Docker image too for consistency.
    - We don't need this `.egg-info`. It gets created because `middleman-protocol` gets installed with `pip --editable`.
    - Let's switch to installing it with `setup.py install`
    - Let's install it with `setup.py install` on the host too. This means you should use `cacheable-requirements` on the host too.
    - After installing `middleman-protocol` in the Docker image with `setup.py install`, remove the `middleman_protocol/` directory you installed from. It won't be used at runtime and having it there would be misleading.

containers/concent-api/Dockerfile.j2
cameel: After discussing this I see what you mean. `install-concent-api.sh` only installs the Python dependencies. In that case we should just change its name to `install-concent-api-python-dependencies.sh` to make it clear.

containers/Makefile
```diff
@@ -83,7 +91,7 @@ endif
 $(info IMAGE_PREFIX=$(IMAGE_PREFIX))
 
 
-all: nginx-proxy nginx-storage concent-api postgresql verifier signing-service concent-signing-service-package geth
+all: nginx-storage concent-api postgresql verifier signing-service concent-signing-service-package geth nginx-proxy
```
cameel: As discussed today - the order should not affect the build result. If it does, the problem is elsewhere.

From what you said it seems that the problem here is that building `nginx-proxy` results in `RELEASE-VERSION` being left in Concent source. Removing it would solve the problem.

containers/concent-api/install-middleman-protocol.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+
+virtualenv_dir=/srv/http/virtualenv
+middleman_protocol_dir=/srv/http/middleman_protocol/
+source "$virtualenv_dir/bin/activate"
+
+cd "$middleman_protocol_dir"
+python setup.py install_lib
```
cameel: Please add a comment that explains why you're using `install_lib` rather than `install`.

containers/signing-service/Dockerfile.j2
cameel: Because it depends on it.

But I see that `install-signing-service.sh` mostly installs its dependencies. In that case it makes sense for it to be before but perhaps we should rename the script it make it clear.

kubernetes/run-unit-tests.sh
```diff
@@ -1,10 +1,20 @@
...
 # Clear the default `local_settings.py` shipped with the container.
 # We're not using it but Django loads it anyway and crashes because it tries to
 # import settings from a config map which is not present here.
 echo -n "" > "/srv/http/concent_api/concent_api/settings/local_settings.py"
 
+(
+    source "$virtualenv_dir/bin/activate"
+    cd /srv/http/middleman_protocol/
+    python -m pip uninstall middleman_protocol
+    rm --recursive /srv/http/virtualenv/lib/python3.6/site-packages/middleman_protocol/
```
cameel: You have python version hard-coded here.

PR #318, opened 2019-01-28
Make ansible playbooks pass lists of packages to apt/pip instead of using loops

cameel:


cameel:
You mean the calls which install from the `requirements` files?

```
        - name:   Install requirements for golem
          pip:
            requirements:       "{{ golem_dir }}/{{ item }}"
            virtualenv:         ~/golem_virtualenv
            virtualenv_python:  "python{{ python_version }}"
          with_items:
            - requirements.txt
            - requirements-test.txt
```

It's not the same. It does not trigger any deprecation warnings. I don't even know if it works with lists. I think it's better to leave it as is.

cameel:
Note: this has actually been merged into `master` (ec34ecc5cda4038543f4cf0ad7a2d5f6b98c4ac2) but github did not notice because I pushed `master` and removed the branch in the same operation.

Issue #322, opened 2019-01-30
Concent VM: Ubuntu with Vagrant 2.2.1 path problem


rwrzesien:
There is a problem on Ubuntu with Vagrant 2.2.1 that `golem_configuration_directory` requires absolute path.

PR #323, opened 2019-01-30
Bugfix change golem configuration directory to use absolute path

rwrzesien:
Resolves https://github.com/golemfactory/concent-deployment/issues/322.

### DIFF
--- a/concent-vm/Vagrantfile
+++ b/concent-vm/Vagrantfile
@@ -12,7 +12,8 @@ Vagrant.configure("2") do |config|
 
     config.vm.network "private_network", ip: "172.40.2.3"
 
-    golem_configuration_directory = './disk/golem_configuration_directory.vdi'
+    # Path must be absolute to work correctly on Ubuntu with Vagrant 2.2.1.
+    golem_configuration_directory = File.expand_path('./disk/golem_configuration_directory.vdi')
 
     config.vm.provider "virtualbox" do |virtualbox|
         virtualbox.memory = 2048


cameel:
@rwrzesien You have your fix mixed in with the workaround for missing Python 3.6 made @bartoszbetka and the branch that adds Golem GUI to the machine. Please put your change directly on master and use that in the pull request.

rwrzesien:
@cameel done :+1:

concent-vm/Vagrantfile
```diff
@@ -12,7 +12,8 @@ Vagrant.configure("2") do |config|
 
     config.vm.network "private_network", ip: "172.40.2.3"
 
-    golem_configuration_directory = './disk/golem_configuration_directory.vdi'
+    # Path must be absolute to work correctly on Ubuntu with Vagrant 2.2.1.
+    golem_configuration_directory = File.expand_path('./disk/golem_configuration_directory.vdi')
```
cameel: I'd rephrase this comment because now it sounds like "it doesn't work on Ubuntu with Vagrant 2.2.1 but I won't tell you why" :)
``` python
# This should be an absolute path because `vboxmanage` fails when used with
# a relative one on some systems (discovered on Ubuntu xx.xx with Vagrant 2.2.1).
```

cameel: @rwrzesien What about my remark here?

rwrzesien: I think I must have missed it, should I add it now?

cameel: I have already added that change in #325. You could review it :)

PR #324, opened 2019-02-01
Bugfix add installation of python3.6 using pyenv after it was deleted from debian testing repository

### LINKED ISSUE #321: Python 3.6  has been removed from debian testing repository
Python `3.6` package  has been removed from debian `testing` repository and currently it's can be find only in `unstable` repository. We must find some solution to this problem.

According information:
https://tracker.debian.org/news/1025205/python36-removed-from-testing/
https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=python3.6

bartoszbetka:
Resolves #321


containers/verifier/build-dependencies.txt
```diff
@@ -0,0 +1,16 @@
+libssl-dev
+git
+pkg-config
+libffi-dev
+software-properties-common
```
cameel: Could we have a separate file just for pyenv/python dependencies that's the same in every container? It would make managing these dependencies easier.

containers/verifier/install-dependencies.sh.j2
```diff
@@ -70,10 +76,6 @@ chmod 555 /usr/local/bin/su-exec
 su-exec nobody true
 
 # Python packages needed by Golem's imgverifier and Concent's verifier
-python3.6 -m pip install --upgrade pip
-python3.6 -m pip install --no-cache-dir --requirement /golem/work/imgverifier-requirements.txt
-
-# Clean up
-apt-get clean
-apt autoremove --assume-yes
-rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* ~/.cache/pip/*
```
cameel: This needs to be done inside the script. Otherwise it'll remain in the image.

concent-vm/consts.yml
```diff
@@ -3,3 +3,6 @@ concent_dir:                 /home/vagrant/concent
 golem_dir:                   /home/vagrant/golem
 golem_hyperdrive_repository: https://github.com/mfranciszkiewicz/golem-hyperdrive/releases/download/v0.2.6/hyperg_0.2.6_linux-x64.tar.gz
 golem_hyperdrive_checksum:   sha256:94220ef8533030652d4a1c126f06fa9cdce9d620b91f1c4574a109a3891e8adf
```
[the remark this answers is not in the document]
bartoszbetka: These lines don't change in this pull request

containers/verifier/Dockerfile.j2
```diff
@@ -10,6 +11,7 @@ RUN  /tmp/install-dependencies.sh
 COPY requirements.lock              /srv/http/concent_api/
 COPY middleman_protocol             /srv/http/middleman_protocol/
 COPY install-concent-api.sh         /tmp/
+COPY clean-container.sh             /tmp/
```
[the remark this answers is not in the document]
bartoszbetka: This is your solution that is already implemented in `signing-service` container. Why you think that it can't remove the files ?

concent-vm/consts.yml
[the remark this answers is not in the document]
cameel: I know. You can do it in a separate pull request. It must have slipped through the review earlier somehow and I only noticed it now.

containers/verifier/Dockerfile.j2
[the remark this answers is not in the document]
cameel: You're right, sorry. I thought I saw `RUN` but no I see it's just `COPY` and the script is executed by another script. I must have been distracted by someone when I looked at it.

containers/verifier/install-dependencies.sh.j2
cameel: No, this one's different. `Dockerfile` runs `install-dependencies.sh` directly which means that once it ends, everything that was not removed by it remains in the image layer that the `RUN` command creates.

If you don't uninstall build dependencies here, they'll remain in the image.

concent-builder/common_tasks/postrequire-configure-tasks.yml
```diff
@@ -1,14 +1,39 @@
...
+    - llvm
+    - libncurses5-dev
+    - libncursesw5-dev
+    - xz-utils
+    - tk-dev
+
+- name:   Install python3.6 package
+  shell: "{{ pyenv_dir }}/bin/pyenv install {{ specific_python_version_for_penv }}"
+  environment:
+    PYENV_ROOT: "{{ pyenv_dir }}"
```
cameel: This fails if you run it twice (e.g. by running `vagrant provision` to update the machine.

```
TASK [Install python3.6 package] ***********************************************
fatal: [concent-vm]: FAILED! => {
    "changed": true,
    "cmd": "/opt/pyenv/bin/pyenv install 3.6.8",
    "delta": "0:00:00.017523",
    "end": "2019-02-04 15:52:08.050721",
    "rc": 1,
    "start": "2019-02-04 15:52:08.033198"
}

STDERR:

pyenv: /opt/pyenv/versions/3.6.8 already exists


MSG:

non-zero return code
```

And anyway, Ansible tasks are supposed to be idempotent. It should not run if Python is already installed. You could use `changed_when` for that purpose. See [Error Handling In Playbooks > Overriding The Changed Result](https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#overriding-the-changed-result).

cameel: Alternatively, you could tell `pyenv` to do nothing if the expected version is installed:
``` bash
pyenv install 3.6.8 --skip-existing
```

concent-builder/common_tasks/postrequire-configure-tasks.yml
```diff
@@ -1,14 +1,39 @@
...
+    - libreadline-dev
+    - libsqlite3-dev
+    - llvm
+    - libncurses5-dev
+    - libncursesw5-dev
+    - xz-utils
+    - tk-dev
+
+- name:   Install python3.6 package
+  shell: "{{ pyenv_dir }}/bin/pyenv install {{ specific_python_version_for_penv }}"
```
cameel: Why put the whole command in quotes? Wouldn't this make more sense:

``` yaml
  shell: "{{ pyenv_dir }}/bin/pyenv" install "{{ specific_python_version_for_penv }}"
```

bartoszbetka: This is not a valid yaml format

cameel: Strange. Well, I meant this by analogy to something like this:

``` yaml
file:   "path={{ item }}" state=absent
```
but I see the quote at the start of the line trips it up. It only accepts this (which I prefer anyway):
``` yaml
file:   path="{{ item }}" state=absent
```

Let's leave it as is in that case and hope that `pyenv_dir` never contains spaces.

containers/concent-api/build-dependencies.txt
```diff
@@ -0,0 +1,19 @@
...
+libssl-dev
+zlib1g-dev
+libbz2-dev
+libreadline-dev
+libsqlite3-dev
+llvm
+libncurses5-dev
+libncursesw5-dev
+xz-utils
+tk-dev
```
cameel: It would be better to share dependency lists between containers. Let's talk about it.

containers/concent-api/clean-container.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -e
+
+
+readarray temporary_dependencies < ${BASH_SOURCE%/*}/build-dependencies.txt
```
cameel: `clean-container.sh` and `clean-pyenv-dependencies.sh` differ only with the name the the txt file. Please make the name parameter and remove one the scripts.

containers/concent-api/install-concent-api.sh.j2
```diff
@@ -1,15 +1,13 @@
...
+virtualenv --python "$python" "$virtualenv_dir"
 source "$virtualenv_dir/bin/activate"
 python -m pip install --upgrade pip
 # The requirements needs to be installed from the directory it's in because it contains relative paths.
 cd /srv/http/concent_api/
 python -m pip install --no-cache-dir --requirement "/srv/http/concent_api/requirements.lock"
 python -m pip install --no-cache-dir gunicorn=="{{ gunicorn_version }}"
-
-# Clean up
-rm -rf ~/.cache/pip/*
```
cameel: Please add this to `clean-container.sh`.

containers/concent-api/install-dependencies.sh.j2
```diff
@@ -1,24 +1,25 @@
...
     gcc
     libssl-dev
-    git
     pkg-config
-    libsecp256k1-dev
     libffi-dev
     software-properties-common
+
+    # Package used by golem-messages
+    libsecp256k1-dev
```
cameel: containers/concent-api/install-dependencies.sh.j2

containers/concent-api/install-dependencies.sh.j2
```diff
@@ -42,7 +43,8 @@ apt-get install              \
     --no-install-recommends  \
     ${testing_dependencies[*]}
 
-# Clean up
-apt-get clean
-apt autoremove --assume-yes
-rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+# Configure pyenv packet manager and install python package
+${BASH_SOURCE%/*}/configure-pyenv.sh
```
cameel: `packet manager` -> `package manager`

concent-vm/consts.yml
```diff
@@ -3,3 +3,6 @@ concent_dir:                 /home/vagrant/concent
...
+pyenv_repository:            https://github.com/pyenv/pyenv
+pyenv_dir:                   /opt/pyenv
+python_path:                 /opt/pyenv/versions/{{ specific_python_version_for_penv }}/bin/python{{ python_version }}
```
cameel: One more thing. You're still using `specific_python_version_for_penv` here. And the name has a typo (should be `pyenv`).

PR #325, opened 2019-02-04
Fix golem configuration directory creation in Concent VM

cameel:
Fixes a regression introduced in #323.

`vboxmanage` fails if the `disk/` subdirectory does not exist when the command is executed. As a result it's impossible to create the machine from scratch with `vagrant up`.

There's a bit of refactoring in addition to the fix:
- Fixed the comment mentioned in https://github.com/golemfactory/concent-deployment/pull/323/files#r252697488
- Renamed `golem_configuration_directory` (it's not a directory)

### DIFF
--- a/concent-vm/Vagrantfile
+++ b/concent-vm/Vagrantfile
@@ -14,12 +14,15 @@ Vagrant.configure("2") do |config|
 
     # This should be an absolute path because `vboxmanage` fails when used with
     # a relative one on some systems (discovered on Ubuntu xx.xx with Vagrant 2.2.1).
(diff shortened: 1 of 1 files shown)


cameel:
I've pushed a fix. Apparently `expand_path()` does not add the trailing slash. Should work now.

cameel:
The problem with vagrant failing to create the disk is a separate thing. It was already broken on `master`. I have created a separate issue for it: #326. The fix is in #327. It depends on changes made here so I'd prefer to merge this pull request first.

Issue #319, opened 2019-01-28
Concent VM: configure.yml playbook does not install rustc from testing


cameel:
An attempt to install golem in the VM ends with an error caused by `rustc` being too old. The `configure.yml` playbook apparently installs `rustc` 1.24 from the main repositories while `testing` has 1.31. That's weird because we have recently modified the playbook to install it from `testing`: #307.

### Reproducing the problem
I install golem in a fresh VM this way:
``` bash
cd concent-vm/
vagrant up

ansible-playbook install-golem.yml                                 \
    --extra-vars  golem_version=develop                            \
    --private-key .vagrant/machines/default/virtualbox/private_key \
    --user        vagrant                                          \
    --inventory   inventory
```

The output from `install-golem.yml` ends with:
```
TASK [Run setup.py develop] **********************************************************************************************************************************************
fatal: [concent-vm]: FAILED! => {
    "changed": true,
    "cmd": "~/golem_virtualenv/bin/python setup.py develop",
    "delta": "0:00:01.517685",
    "end": "2019-01-28 12:44:31.120656",
    "rc": 1,
    "start": "2019-01-28 12:44:29.602971"
}

STDOUT:

Task Collector already built
running develop
running egg_info
writing golem.egg-info/PKG-INFO
writing dependency_links to golem.egg-info/dependency_links.txt
writing entry points to golem.egg-info/entry_points.txt
writing requirements to golem.egg-info/requires.txt
writing top-level names to golem.egg-info/top_level.txt
reading manifest file 'golem.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'golem.egg-info/SOURCES.txt'
running build_ext
running build_rust


STDERR:

Dynamically determined version: 0.18.3+dev842.g90930bb
warning: no files found matching 'golem/ethereum/genesis_golem.json'
warning: no files found matching 'golem/ethereum/mine_pending_transactions.js'
warning: no files found matching '*.jpg' under directory 'apps'
warning: no files found matching '*.lxs' under directory 'apps'
warning: no files found matching '*.ply' under directory 'apps'
warning: no files found matching '*.lxo' under directory 'apps'
warning: no files found matching '*.lxm' under directory 'apps'
warning: no files found matching '*.lxv' under directory 'apps'
       Fresh ucd-util v0.1.3
       Fresh cfg-if v0.1.6
       Fresh utf8-ranges v1.0.2
       Fresh version_check v0.1.5
[...]

PR #328, opened 2019-02-04
Quick and dirty workaround for rustc not getting installed from testing in Concent VM

cameel:
We don't know the cause and it may take some time until we investigate it properly. We need something to make the machine work in the meantime.

This change should we reverted when #319 is fixed.

### DIFF
--- a/concent-vm/configure.yml
+++ b/concent-vm/configure.yml
@@ -60,6 +60,14 @@
               - python3-psutil
               - rustc
 
+        # FIXME: Why doesn't the task above install rustc from testing? Find the cause and fix it properly.
+        # Reported as https://github.com/golemfactory/concent-deployment/issues/319
+        - name:   Quick and dirty workaround for rustc not getting installed from testing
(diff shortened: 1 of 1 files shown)


concent-vm/configure.yml
```diff
@@ -60,6 +60,14 @@
               - python3-psutil
               - rustc
 
+        # FIXME: Why doesn't the task above install rustc from testing? Find the cause and fix it properly.
+        # Reported as https://github.com/golemfactory/concent-deployment/issues/319
+        - name:   Quick and dirty workaround for rustc not getting installed from testing
+          shell:  apt-get install rustc
```
bartoszbetka: Why you are not using backslash here ?

cameel: Because it's not needed.

See this about line breaking in YAML: [Understanding multi line strings in YAML and Ansible (Part I - YAML)](https://adminswerk.de/multi-line-string-yaml-ansible-I/)

PR #332, opened 2019-02-07
Automate and simplify Concent VM setup

cameel:
This pull request changes the way the VM is created. Now can all be built with a single `vagrant up` command.


concent-vm/Vagrantfile
```diff
@@ -5,20 +5,44 @@ Vagrant.configure("2") do |config|
...
+        config.vm.network "forwarded_port", guest: start_port,     host: start_port
+        config.vm.network "forwarded_port", guest: start_port + 1, host: start_port + 1
+    end
+
     # This should be an absolute path because `vboxmanage` fails when used with
     # a relative one on some systems (discovered on Ubuntu xx.xx with Vagrant 2.2.1).
     disk_dir                  = File.expand_path('./disk/')
     $golem_configuration_disk = "#{disk_dir}/golem_configuration_directory.vdi"
 
+    if ENV.key?('CONCENT_VM_INSTALL_SHOW_GUI')
```
bartoszbetka: What is purpose of this statement ? This environment variable is also undefined in readme. When the `CONCENT_VM_SHOW_GUI` environment variable is set to false, the virtualbox gui will be run anyway by `CONCENT_VM_INSTALL_GOLEM_ELECTRON` . I think that only the `CONCENT_VM_SHOW_GUI` environment should be decide about running virtualbox gui.

Please also consider delete `CONCENT_VM_SHOW_GUI` and decide to run virtualbox gui by installation of golem electron(`CONCENT_VM_INSTALL_GOLEM_ELECTRON` set to True or False).

cameel: The purpose is to allow showing/hiding VM screen independently of `CONCENT_VM_INSTALL_GOLEM_ELECTRON`.

By default it works like you described: when `CONCENT_VM_SHOW_GUI` is not defined, screen is shown if you have `golem-electron` and hidden if you don't. You don't have to think about it.

But what if you have GUI installed but you're working on a feature that does not require it and it gets in the way? You can define `CONCENT_VM_SHOW_GUI=false` hide it, overriding the default. Or if you don't have `golem-electron` but you installed a desktop environment on your own or you want to access the TTY? You can still show the screen with `CONCENT_VM_SHOW_GUI=true`.

> This environment variable is also undefined in readme.
Yeah, it's not. I'll add a description.

cameel: I have added a README section that describes all the variables. Check it out.

cameel: There was actually an error there. `CONCENT_VM_INSTALL_SHOW_GUI` in the condition should be `CONCENT_VM_SHOW_GUI`. Fixed. Now the condition works I described.

PR #336, opened 2019-02-08
Add scripts that generate keys for signing-service

### LINKED ISSUE #335: Automate generating keys for Signing Service
Currently we only have manual instructions for generating keys for the Signing Service. These are keys used by golem-messages and their format is publicly known. We could make it simpler though.

1) Create a Python script (`generate-ecc-key-pair.sh`) that:
    - Generates an ECC key pair
    - Encodes the keys in the format that can be passed to Signing Service
    - Prints the keys to the screen
2) Put the script in `/usr/local/bin` in the `signing-service` container
3) Add a Docker command for running the script to the README and explain when it should be used

bartoszbetka:
Resolves #335


containers/signing-service/generate-ecc-key-pair.py
```diff
@@ -0,0 +1,10 @@
+from golem_messages import ECCx
+from base64 import b64encode
+
+ecc = ECCx(None)
+
+signing_service_pub_key=b64encode(ecc.raw_pubkey).decode('utf-8')
+signing_service_priv_key=b64encode(ecc.raw_privkey).decode('utf-8')
```
cameel: Please use full words: `private`/`public`.

README.md
```diff
@@ -372,6 +372,15 @@ If you have received the package and only want to build and run the docker conta
         concent-signing-service:latest
     ```
 
+### Generate Signing Service keys
+To generate public and private keys for signing-service use the `generate-ecc-key-pair.sh` script.
```
```
### Generating a key pair
Signing Service needs a key pair to secure its communication with Concent.
These are the same keys that are used for signing messages generated by golem-messages.
They're just base64-encoded so that they can be easily passed on the command line or in shell variables.

The public key needs to be known to Concent.
The private one must not - this is the whole point of having a separate service for signing.

When setting up a new instance of the service, you can easily generate a new pair using the `generate-ecc-key-pair.sh` script included in the Docker image:
```

README.md
```diff
@@ -372,6 +372,15 @@ If you have received the package and only want to build and run the docker conta
...
+```bash
+docker run                                  \
+    --interactive                           \
+    --tty                                   \
+    --entrypoint generate-ecc-key-pair.sh   \
+    concent-signing-service
+```
```
cameel: That's not the purpose of `--entrypoint`. It would be better to set `--entrypoint /bin/bash` and run the command as usual. This makes it clear that you simply want to override the default entrypoint.

PR #345, opened 2019-02-13
Fix nginx router regex to corectly mapping golem-messages version headers to properly clusters

bartoszbetka:


kubernetes/config-maps/nginx-proxy/configs/http.conf.j2
```diff
@@ -32,10 +32,10 @@ upstream {{ previous_golem_messages_version }} {
 map $http_x_golem_messages $version {
      # This address uses the HTTP protocol because it is an internal endpoint that is on this cluster.
      ""          "http://{{ current_golem_messages_version }}";
-     "~^{{ current_golem_messages_version }}.*$"   "http://{{ current_golem_messages_version }}";
+     "~^{{ current_golem_messages_version }}\..+$"   "http://{{ current_golem_messages_version }}";
```
cameel: Hmm... this won't work with something like `2.3.4rc1` which is a common convention.

Maybe change it so that instead of expecting the third dot, it does not ignore number that immediately follow the version? Then `2.3.4rc1` would match `2.3.4` but `2.3.45rc1` would not.

The only problem is when `current_golem_messages_version` itself does not end with a number. E.g. it's `2.3.4-pre`. Then `2.3.4-pre2` would be considered incompatible even though it's compatible. But it's unlikely that we'll use something like that. There should be just a comment about it somewhere.

bartoszbetka: You are not right. Nginx used perl regex. It is not expected the third dot. The `dot` in perl regex mean - any single character. The `+` mean match one or unlimited times. We decided that third part can be anything. This regex assumes that third part can be one or more characters, for example (`2.15.*`)

bartoszbetka: We want to change this assumption ?

cameel: > You are not right. Nginx used perl regex. It is not expected the third dot. The `dot` in perl regex mean - any single character. The `+` mean match one or unlimited times.

I know that. I was just assuming that `current_golem_messages_version` might contain a version composed of more than two numbers. E.g. `1.2.3`.

> We want to change this assumption ?

No. I did not remember if we made this assumption. If we did, then your code is fine.

We should change names as suggested in #346, to make this assumption more explicit.

Please also add comments explaining what is allowed and what is  not in these variables.

PR #354, opened 2019-02-27
Feature encrypting secrets at rest with using ansible vault

### LINKED ISSUE #206: Encrypting secrets at rest with Ansible Vault
This task depends on #204 and #205.

### Overview
There are four types of places where we store secrets:
1. On the machines of people who run Ansible playbooks to provision the build server.
2. On the build server in a global directory.
3. On the build server in a local build directory (briefly, when Kubernetes secret manifests are built).
4. On the cluster.

In all of these places secrets are currently unencrypted. Ultimately they must end up unencrypted on the cluster for the application to be able to access them but we could secure them in the other locations to decrease the chances of them leaking.

The goal of this task is to use [Ansible Vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html) to encrypt our secrets in locations 1. and 2. and decrypt them just before the Kubernetes secret manifests are built.

### Secret encryption
#205 shows how our secrets are structured now. They should still have that structure when decrypted.

When encrypted, the directories should stay the same, but each file should get an additional `.vault` extension. E.g. `concent-secrets/concent-dev/cluster-secrets.yml` should become `concent-secrets/concent-dev/cluster-secrets.yml.vault` when encrypted.

#### Vault passwords
Vault password is the same for all encrypted files within one "secret set" (e.g. all secrets for `concent-dev` or `cluster`). But different sets may have different passwords. Currently we need three passwords:
1. One for `concent-dev`, `concent-storage` and `concent-testnet`. These are not very sensitive and the password can be weak and convenient.
2. One for `concent-mainnet`. This is highly sensitive and needs a strong password.
3. One for `cluster`. This one gives access to the whole cloud infrastructure and should have a strong password too. With access to the infrastructure it's easy to intercept mainnet secrets in an unencrypted form.

The passwords should be decided ahead of time and shared between all the people who have permission to manage the cloud infrastructure. People that only deploy the application to a single cluster **do not** need to know these passwords. They can expect that the secrets they need are either already on the cluster (if needed by the application) or on the build server (if needed for deployment). 

Each person with access to the password is responsible for securing it by either memorizing it and never writing it down or using a password manager.

### Uploading and decrypting secrets
Secrets should never be stored outside of the deployment server or the cluster in an unencrypted form. 

Upload and decryption should be performed by the `cluster-deploy-secrets.yml` playbook using Ansible's `copy` module. The playbook should ask user for passwords to them (Ansible can do that automatically). Then it should decrypt the secrets and put them in in the build directory. In that form they can be used by `secrets/Makefile`.

Secrets (and all files that contain them) must always be removed after the build and deployment, no matter whether the operation is successful or not.

### Clean up unencrypted secrets
After you implement this, make sure that there are no unencrypted secrets left on the build server. Tell other people to delete them from their machines too.

Then change the master password to Cloud SQL.

We don't have `mainnet` secrets yet and those on the other clusters are not sensitive so changing them is not necessary. Still, it might be a good idea to change `testnet` passwords.

bartoszbetka:
Resolves #206


README.md
```diff
@@ -17,11 +17,14 @@ gcloud compute disks create --size 30GB <disk name>
...
 This operation requires privileges for creating and deleting arbitrary databases in Cloud SQL.
 For security reasons the role used to access the database from within the cluster should not have such wide privileges and this step needs to be performed outside of cluster deployment process.
+This step require knowledge of cloud and cluster passwords to begin decrypting process.
 
 ``` bash
 cd concent-deployment/cloud/
 ansible-playbook create-databases.yml                              \
+    --vault-id @prompt                                             \
+    --ask-vault-pass                                               \
```
cameel: From the docs I see that both `--vault-id @prompt` and `--ask-vault-pass` do the same thing. Let's settle on one and use it consistently. `--vault-id` seems to be newer and is now recommended so let's use it in all the commands instead of `--ask-vault-pass`.

cloud/configure-user-authentication-for-clusters.yml
```diff
@@ -0,0 +1,30 @@
...
+        - name: Create a directory for secrets
+          file:
+            path:  "{{ data_dir }}/cloud-secrets/"
+            state: directory
+
+        - name: Upload decrypted "{{ gke.service_account_name }}-private-key.json" secret for activate google cloud service account
+          copy:
+            src:  "{{ local_secret_dir }}/cloud/{{ gke.service_account_name }}-private-key.json"
+            dest: "{{ data_dir }}/cloud-secrets/{{ gke.service_account_name }}-private-key.json"
+            mode: 0444
```
cameel: Let's upload the key to user's home directory. No need to put it in `{{ data_dir }}` if we're not storing it there permanently.

README.md
```diff
@@ -153,13 +156,16 @@ Do this if you want to use the remote server for building and deploying.
     Where the `$user` shell variable contains the name of your shell account on the remote machine.
 
 
-### Uploading secrets and access to deploy on specific gcloud environment
-This is a separate step that is needed on the new server and in the case of changes in secrets
+### Access to deploy on specific gcloud environment
+This is a separate step that is needed for each linux user on the server that want to access to deploy on the clusters.
+The `$user_name` variable controls which user gets permission for deploying process.
+This step require knowledge of cloud password to begin decrypting process.
```
```
### Authorizing a user account on the build server to access the clusters
This step must be performed separately for every user of the build server who needs to be able to access other parts of the project infrastructure on Google Cloud with `kubectl` or `gcloud`.
It can be performed by user himself or an admin who can impersonate him with `sudo`.

The `$user_name` variable below indicates the user account to be authorized.
To perform this step you need to have the .vault files with encrypted secrets in your local `concent-secrets/` directory.
Only cloud secrets are required in this case.
Ansible will prompt you for password required to decrypt them.
```

README.md
```diff
@@ -188,12 +194,14 @@ ansible-playbook build-test-and-push.yml                           \
 
 ### Deploying secrets
 Before you can deploy containers, you need to make sure that certificates, keys and passwords used to configure those containers are available on the cluster.
+This step require knowledge of cluster password to begin decrypting process.
```

``` markdown
To perform this step you need to have the .vault files with encrypted secrets in your local `concent-secrets/` directory.
Only cluster secrets are required in this case.
Ansible will prompt you for password required to decrypt them.
```

cameel:
Please remember about the last part of the task before closing the issue: change passwords, share them with other team members and make sure that there are no unencrypted keys left anywhere.

cloud/cluster-deploy-secrets.yml
```diff
@@ -7,11 +7,27 @@
...
+          file:
+            path:  "{{ build_dir }}/concent-deployment/secrets/{{ cluster }}"
+            state: directory
+
+        - name: Upload decrypted secrets for build secrets purpose
+          copy:
+            src:     "{{ local_secret_dir }}/{{ cluster }}/{{ item }}"
+            dest:    "{{ build_dir }}/concent-deployment/secrets/{{ cluster }}"
+          with_items:
+            - cluster-secrets.yml
```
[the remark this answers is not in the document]
bartoszbetka: Ok. I modify soultion to use `.vault` extension.

README.md
bartoszbetka: This playbook use two different password for cloud and cluster secrets. That situation in ansible not work with one flag.

cloud/configure-user-authentication-for-clusters.yml
bartoszbetka: Do you assume that home directory always exists ?

PR #360, opened 2019-03-13
Chore concent vm with a single configuration file

### LINKED ISSUE #278: Concent VM: a single configuration file
Currently setting up the virtual machine requires creating two configuration files (`extra_settings.py` and `signing-service-env.sh`) and a shell variable (`CONCENT_DEPLOYMENT_VERSION`). There will be more as we make the machine more versatile (e.g. `app_cfg.ini`) for Golem. The goal of this task is replace it all with a single configuration file.

1) The configuration file should be called `config.yml`.
    - It should not be stored in the repository (add to `.gitignore`).
    - If the file does not exist, `Vagrantfile` should exit with a clear error message.
2) `extra_settings.py` should now be created automatically inside the machine from a template and contain all settings from a dict called `concent_local_settings`.
3) `signing_service-env.sh` should now be created automatically inside the machine from a template and contain all shell variables from a dict called `signing_service_env`.
4) `CONCENT_DEPLOYMENT_VERSION` should be replaced with a top-level variable called `concent_deployment_version`. The playbooks can read it directly from `config.yml` instead of expecting Vagrant to pass it to them.

#### `config.yml`
``` yaml
concent_deployment_version: dev

concent_local_settings:
    USE_SIGNING_SERVICE:          true
    ADDITIONAL_VERIFICATION_COST: 1
    MIDDLEMAN_ADDRESS:            middleman.default.svc.cluster.local
    MIDDLEMAN_PORT:               9054

signing_service_env:
    ETHEREUM_PRIVATE_KEY:        "ZjIwODU2ZmFhN2YxOTQzN2E0MGIwODAxYjdlNDc1Y2U0ZWZhNTkzMmVkYmFhYmMyMjFiODQ5N2VlMGM1YzRiNSAgLQo="
    SIGNING_SERVICE_PRIVATE_KEY: "OTYyNmMxYzJhZmJlYzVjMmM5NGNlYmRjMzE4N2M5MjcgIC0K"
```

#### `extra_settings.py`
The generated file should contain only the following:
``` python
USE_SIGNING_SERVICE = True
ADDITIONAL_VERIFICATION_COST = 1
MIDDLEMAN_ADDRESS = 'middleman.default.svc.cluster.local'
MIDDLEMAN_PORT = 9054
```

#### `signing_service-env.sh`
The generated file should contain only the following:
``` bash
export ETHEREUM_PRIVATE_KEY="ZjIwODU2ZmFhN2YxOTQzN2E0MGIwODAxYjdlNDc1Y2U0ZWZhNTkzMmVkYmFhYmMyMjFiODQ5N2VlMGM1YzRiNSAgLQo="
export SIGNING_SERVICE_PRIVATE_KEY="OTYyNmMxYzJhZmJlYzVjMmM5NGNlYmRjMzE4N2M5MjcgIC0K"
```

bartoszbetka:
Resolves #278


concent-vm/Vagrantfile
```diff
@@ -140,7 +140,15 @@ Vagrant.configure("2") do |config|
         ansible.limit              = "all"
     end
 
+    $concent_config_file="config.yml"
```
cameel: This does not need to be a global variable.

cameel: Let's call it `concent_vm_config_file`.

concent-vm/Vagrantfile
```diff
@@ -140,7 +140,15 @@ Vagrant.configure("2") do |config|
...
     if ENV['CONCENT_VM_INSTALL_CONCENT'] == 'true'
+        if not File.exists?($concent_config_file)
+          puts "The installation of concent in the virtual machine require a file called config.yml." +
+               " This file contain imported settings and secrets for concent and signing-service." +
+               " Please create it and rerun vagrant."
```
cameel: Please reword:
```
"Installation of Concent in the virtual machine requires a file called #{concent_config_file}. "
"This file should contain values that will be used in Concent's local_settings.py and as command-line parameters for the Signing Service. "
"Please create it and rerun vagrant."
```

concent-vm/configure.yml
```diff
@@ -120,6 +120,7 @@
             src:        "{{ concent_deployment_dir }}/concent-vm/nginx-storage.service"
             dest:       "/etc/systemd/system/nginx-storage.service"
             remote_src: yes
+            mode:       0555
```
cameel: Your commit description is too vague (Fix the problem with `nginx-storage.service`). It should say more about the problem. What exactly are you fixing?

Also, I think it should be a separate pull request because it's completely unrelated.

cameel:
Oh, actually one more thing about the config: README needs to be updated.

concent-vm/configure.yml
bartoszbetka: Ok, I create separate pull request for this.

PR #361, opened 2019-03-15
Modify `concent-verifier` docker image to use Golem's new `blender_verifier` as a base image

bartoszbetka:


containers/Makefile
```diff
@@ -196,23 +191,13 @@ $(SIGNING_SERVICE_SOURCE_FROM_CONCENT_API): build/signing-service/%: build/conce
...
 	cp "$<" "$@"
 
-build/verifier/golem/imgverifier-requirements.txt: build/repositories/golem/apps/rendering/resources/images/requirements.txt
-	cp "$<" "$@"
-
 build/repositories/concent: build/package-builder/clone-or-update.sh /var/concent-data/concent
 	build/package-builder/clone-or-update.sh "concent" "$@"
 
-build/repositories/golem/: build/package-builder/clone-or-update.sh /var/concent-data/golem
-	build/package-builder/clone-or-update.sh  "golem" "$@"
```
cameel: `clone-or-update.sh` still has code to support the `golem` argument but will no longer work because `golem_version` was removed.

Please remove the first parameter and rename the script to `clone-or-update-concent.sh`.

containers/versions.yml
```diff
@@ -1,26 +1,25 @@
...
+nginx_debian_version:                  "1.10.3-1+deb9u2"
+openresty_alpine_fat_version:          "1.13.6.2-0"
+postgres_alpine_version:               "10.4"
+postgres_version:                      "9.6"
+debian_postgres_version:               "9.6"
+kubectl_version:                       "1.13.3-00"
+python_version:                        "3.6"
+python_patch_version:                  "8"
+rabbitmq_alpine_version:               "3.7.6"
+su_exec_hash:                          "60e8c3010aaa85f5d919448d082ecdf6e8b75a1c"
```
cameel: `su_exec_hash` is no longer used.

containers/verifier/install-dependencies.sh.j2
```diff
@@ -1,44 +1,18 @@
...
-    xz-utils
-    zlib1g-dev
-)
-
-testing_dependencies=(
-    # Packages needed to build python extensions, dependencies of golem-messages in particular
-    # They need to be installed from testing, because on Debian Jessie they are too old or non-existent
-    binutils
-    build-essential
+    gcc
```
cameel: What are we using `gcc` for?

cameel: I assume that `gcc` is only needed when we're building the image? It would be best to remove it afterwards.

containers/verifier/entrypoint.sh
```diff
@@ -1,3 +0,0 @@
-#!/bin/bash -e
-
-exec /usr/local/bin/su-exec nobody /bin/bash -c "$@"
```
cameel: Are you removing this because it no longer works with the new image or because you noticed that using `USER nobody` in the `Dockerfile` is enough? If it's the latter, then this should really be a separate commit because it has nothing to do with the image change.

containers/concent-api/install-concent-api-python-dependencies.sh.j2
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -e
 
-python="/opt/pyenv/versions/{{ python_version }}.{{ python_patch_version }}/bin/python{{ python_version }}"
+DEFAULT_PYTHON_PATH="/opt/pyenv/versions/{{ python_version }}.{{ python_patch_version }}/bin/python{{ python_version }}"
```
cameel: Is `DEFAULT_PYTHON_PATH` just your own local variable or something global that other programs (e.g. pyenv or virtualenv) expect? If it's the former, please use lower case.

Also, I think that calling it `default_python` would make it clearer that it's meant to be just a default value for the `python` parameter.

containers/verifier/Dockerfile.j2
```diff
@@ -1,31 +1,22 @@
...
-COPY clean-pyenv-dependencies.sh              /tmp/
-COPY golem/imgverifier-requirements.txt       /golem/work/
 COPY install-dependencies.sh                  /tmp/
 RUN  /tmp/install-dependencies.sh
 
 # Build virtualenv and concent apps
 COPY cacheable-requirements.lock                       /srv/http/concent_api/
 COPY install-concent-api-python-dependencies.sh        /tmp/
-RUN  /tmp/install-concent-api-python-dependencies.sh
+RUN  /tmp/install-concent-api-python-dependencies.sh   python3.6
```
cameel: This should be in `containers/versions.py`, rather than hard-coded here.

containers/concent-api/install-concent-api-python-dependencies.sh.j2
cameel: I'm not sure we really need this default. In `Dockerfile` you're passing `python3.6` so the default is not used. Is there any situation in which we do use this default and it's actually useful?

containers/verifier/install-dependencies.sh.j2
bartoszbetka: I know about it. I want to separate packages needed only for building images in the separate pull request. It's not that simple. I write issue for that #364.

bartoszbetka: The same as https://github.com/golemfactory/concent-deployment/pull/361#discussion_r266460697

containers/verifier/entrypoint.sh
bartoszbetka: This was deleted because new the `blender_verfier` docker image has already `entrypoint.sh`. I investigated [entrypoint](https://github.com/golemfactory/golem/blob/develop/apps/entrypoint.sh) and to works with it, we need add the `LOCAL_USER_ID` environmental variable. We could also use `USER` directive to get the same behavior.

containers/concent-api/install-concent-api-python-dependencies.sh.j2
bartoszbetka: Yes, we use default in other docker images. The `verifier` image uses `python3.6` package from system source repository after its based on the Ubuntu distribution.

PR #367, opened 2019-03-20
Change nginx-storage systemd service to run as root and add improvement to it

bartoszbetka:
Running `nginx-storage` systemd service as builder user causes error
code `127` for `run-nginx-storage.sh` script. This error code is define as
`command not found`.


bartoszbetka:
merged

PR #373, opened 2019-03-22
Add ability to control amount of application instances

bartoszbetka:
The solution works quite well, but its design does not look good. Jinja2 has a problem with checking the nested dictionary when its parent may not exist. Available solution: https://stackoverflow.com/questions/48283987/how-to-check-whether-variable-is-defined-when-dictionary-may-be-missing-in-ansi.

### DIFF
--- a/kubernetes/services/concent-api-worker.yml.j2
+++ b/kubernetes/services/concent-api-worker.yml.j2
@@ -3,8 +3,8 @@ kind:       Deployment
 metadata:
   name: concent-api-worker
 spec:
-  replicas:             3
-  revisionHistoryLimit: 3
+  replicas:             {{ concent_versions[concent_version] | chained_get('replica_count.concent_api_worker') | default('{{ replica_count.concent_api_worker }}', true) }}
+  revisionHistoryLimit: 2
   template:
     metadata:
       labels:
--- a/kubernetes/services/concent-api-worker.yml.j2
+++ b/kubernetes/services/concent-api-worker.yml.j2
@@ -19,11 +19,11 @@ spec:
             - containerPort: 80
           resources:
             requests:
-              memory: "{{ resource_limits.worker_apps.concent_api_worker.memory.requests }}"
-              cpu:    "{{ resource_limits.worker_apps.concent_api_worker.cpu.requests }}"
+              memory: "{{ concent_versions[concent_version] | chained_get('resource_limits.worker_apps.concent_api_worker.memory.requests') | default('{{ resource_limits.worker_apps.concent_api_worker.memory.requests }}', true) }}"
+              cpu:    "{{ concent_versions[concent_version] | chained_get('resource_limits.worker_apps.concent_api_worker.cpu.requests') | default('{{ resource_limits.worker_apps.concent_api_worker.cpu.requests }}', true) }}"
             limits:
-              memory: "{{ resource_limits.worker_apps.concent_api_worker.memory.limits }}"
-              cpu:    "{{ resource_limits.worker_apps.concent_api_worker.cpu.limits }}"
+              memory: "{{ concent_versions[concent_version] | chained_get('resource_limits.worker_apps.concent_api_worker.memory.limits') | default('{{ resource_limits.worker_apps.concent_api_worker.memory.limits }}', true) }}"
+              cpu:    "{{ concent_versions[concent_version] | chained_get('resource_limits.worker_apps.concent_api_worker.cpu.limits') | default('{{ resource_limits.worker_apps.concent_api_worker.cpu.limits }}', true) }}"
           volumeMounts:
             - name:      concent-api-secrets
               mountPath: /srv/http/concent_api/concent_api/settings/secrets/
--- a/tools/jinja2-filters.py
+++ b/tools/jinja2-filters.py
@@ -3,18 +3,43 @@ from base64 import b64decode
 class NotAString(Exception):
     pass
 
+class NotADict(Exception):
+    pass
+
+class KeyNotFound(Exception):
+    pass
+
 class UnsupportedDataType(Exception):
     pass
 
-def filter_b64decode(encoded_value, return_type = "string"):
+def filter_b64decode(encoded_value, return_type="string"):
     if not isinstance(encoded_value, str):
-        raise NotAString("The encoded value does not a string")
+        raise NotAString("The encoded value is not a string")
     if return_type == "string":
         # If the input is invalid, decoding can result in the following exceptions:
         # - binascii.Error     - if the input is not base64 encoded
         # - UnicodeDecodeError - if the encoded value is not a valid UTF-8 string
-        return b64decode(encoded_value, validate = True).decode("utf-8")
+        return b64decode(encoded_value, validate=True).decode("utf-8")
     if return_type == "bytes":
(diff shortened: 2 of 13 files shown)


cameel:
Yeah this is really ugly. It would be fine if you needed if just once but you're using it a lot and the code looks insane :) I think the best you can do to make it readable is to create a custom filter. Something like this:

```
{{ concent_versions[concent_version] | chained_get("resource_limits", "worker_apps", "concent_api_worker", "memory", "requests") }}
```
or (if you prefer):
```
{{ concent_versions[concent_version] | chained_get("resource_limits.worker_apps.concent_api_worker.memory.requests") }}
```
this should return the same result as:
```
(
    (
        (
            (
                concent_versions[concent_version].resource_limits | default({})
            ).worker_apps | default({})
        ).concent_api_worker | default({})
    ).memory | default({})
).requests | default('{{ resource_limits.worker_apps.concent_api_worker.memory.requests }}', true)
```

The filter implementation would be very simple. We already have one filter you created before (`filter_b64decode`) - use it as a base (BTW that old filter is no longer used; remove it). Your new filter only needs to loop over key names and in each iteration replace your `value` with `value[key]`.

kubernetes/services/concent-api-worker.yml.j2
```diff
@@ -3,8 +3,8 @@ kind:       Deployment
 metadata:
   name: concent-api-worker
 spec:
-  replicas:             3
-  revisionHistoryLimit: 3
+  replicas:             {{ (concent_versions[concent_version].replicas_count | default({})).concent_api_worker | default('{{ replicas_count.concent_api_worker }}', true) }}
```
cameel: `replicas_count` -> `replica_count`

kubernetes/services/concent-api-worker.yml.j2
```diff
@@ -19,11 +19,27 @@ spec:
             - containerPort: 80
           resources:
             requests:
-              memory: "{{ resource_limits.worker_apps.concent_api_worker.memory.requests }}"
-              cpu:    "{{ resource_limits.worker_apps.concent_api_worker.cpu.requests }}"
+              memory: "{{
+                        ((((concent_versions[concent_version].resource_limits | default({})).worker_apps
+                        | default({})).concent_api_worker | default({})).memory | default({})).requests
+                        | default('{{ resource_limits.worker_apps.concent_api_worker.memory.requests }}', true)
```
cameel: I think that even getting rid of defaults would be better than this. I.e. (if the idea with a filter does not work out), just define these values separately for every cluster in its config instead of inheriting them from `var.yml`.

tools/jinja2-filters.py
```diff
@@ -18,3 +21,19 @@ def filter_b64decode(encoded_value, return_type = "string"):
         return b64decode(encoded_value, validate = True)
     else:
         raise UnsupportedDataType("The encoded value is an unsupported data type")
+
+def filter_chained_get(dictionary_name, chain):
```
cameel: `dictionary_name` -> `dictionary`

tools/jinja2-filters.py
```diff
@@ -18,3 +21,19 @@ def filter_b64decode(encoded_value, return_type = "string"):
...
+        raise NotADict("The variable does not a dict")
+    if not isinstance(chain, str):
+        raise NotAString("The chain segment does not a string")
+
+    chain_segments = chain.split('.')
+    for chain_segment in chain_segments:
+        if chain_segment in dictionary_name.keys():
+            dictionary_name = dictionary_name[chain_segment]
+            if dictionary_name is None:
+                return ("{{" + chain + "}}")
```
cameel: Instead of overwriting `dictionary_name` (which is a parameter) please create a local variable. You can call it `value`.

tools/jinja2-filters.py
```diff
@@ -18,3 +21,19 @@ def filter_b64decode(encoded_value, return_type = "string"):
         return b64decode(encoded_value, validate = True)
     else:
         raise UnsupportedDataType("The encoded value is an unsupported data type")
+
+def filter_chained_get(dictionary_name, chain):
+    if not isinstance(dictionary_name, dict):
+        raise NotADict("The variable does not a dict")
```
cameel: This does not do what I meant. Here's how it should work in edge cases:

Let's assume that `config_values` is a dict that looks like this:
``` yaml
a:
  b:
    d: 2
    e: "string" 
    f:
    g: {}
  c:
```

The expected results are:
- `config_values | changed_get("a.b.d")` = `2`
- `config_values | changed_get("a.b.d.x")` - error
- `config_values | changed_get("a.b.f")` - `null`
- `config_values | changed_get("a.b.f.x")` - `null`
- `config_values | changed_get("a.b.f.g")` - `{}`
- `config_values | changed_get("a.b.f.g.x")` - `null`
- `config_values | changed_get("a.b")` - `{'d': 2, 'e': "string", 'f': None, 'g': {}}`

tools/jinja2-filters.py
```diff
@@ -18,3 +21,19 @@ def filter_b64decode(encoded_value, return_type = "string"):
...
+    if not isinstance(chain, str):
+        raise NotAString("The chain segment does not a string")
```
cameel: These error messages sound weird. Please fix them.

tools/jinja2-filters.py
```diff
@@ -18,3 +24,24 @@ def filter_b64decode(encoded_value, return_type = "string"):
...
+    value = dictionary
+    for chain_segment in chain_segments:
+        if isinstance(value, dict) and chain_segment in value:
+            if value is None:
+                return value
+            value = value[chain_segment]
+        elif isinstance(value, str):
+            raise KeyNotFound("The key value was not found in the dictionary")
+        elif isinstance(value, int):
+            raise KeyNotFound("The key value was not found in the dictionary")
```
cameel: You can do `isinstance(value, (str, int))`

tools/jinja2-filters.py
```diff
@@ -18,3 +24,24 @@ def filter_b64decode(encoded_value, return_type = "string"):
...
+        raise NotADict("The variable is not a dict")
+    if not isinstance(key_list, str):
+        raise NotAString("The chain segment is not a string")
+
+    chain_segments = key_list.split('.')
+    value = dictionary
+    for chain_segment in chain_segments:
+        if isinstance(value, dict) and chain_segment in value:
+            if value is None:
+                return value
```
cameel: This will never be true because if you're here, `value` must be a `dict`, not `None`.

tools/jinja2-filters.py
```diff
@@ -3,12 +3,18 @@
...
+class NotADict(Exception):
+    pass
+
+class KeyNotFound(Exception):
+    pass
+
 class UnsupportedDataType(Exception):
     pass
 
 def filter_b64decode(encoded_value, return_type = "string"):
```
kbeker: There should not be space between equal sign :P

tools/jinja2-filters.py
```diff
@@ -18,3 +24,22 @@ def filter_b64decode(encoded_value, return_type = "string"):
         return b64decode(encoded_value, validate = True)
```
kbeker: Same here :)

PR #379, opened 2019-04-11
Delete uploading and manually activating service accounts after assign them to appropriate google compute engine instances

### LINKED ISSUE #372: Assign our service accounts to Google Compute Engine instances instead of activating them manually
It's best to do this task after #369.

Currently we're managing service account keys along with other secrets. Turns out there is a way to move this responsibility to Google Cloud as long as the account is only used from a GCP instance. The docs say that `gcloud` can automatically use the service account assigned to an instance. This will make it harder to leak these keys.

See:
- [Creating and enabling service accounts for instances > Authenticating tools on an instance using a service account](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#tools)
- [Creating and enabling service accounts for instances > Using the Compute Engine Default Service Account](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#using_the_compute_engine_default_service_account)

1. Update `concent-builder` and `concent-deployment-server` metadata so that they use service accounts we've been activating there manually.
    - This requires a reboot.
    - You may need to also deactivate the accounts we've been using so far or reset `gcloud` settings.
2. Remove `gcloud auth activate-service-account` step from our playbooks.
    - From now on we'll be using the account assigned to an instance.
    - When running these playbooks on a machine outside of GCP, we'll now be using our personal accounts, activated manually.
        - This way the service account credentials will never have to get out of the machine and user accounts of all DevOps members of our team have enough privileges for such an operation and are already accessible on their machines.
        - Note that this is a rare case anyway. It would only be necessary when the **target** machine is outside of Google Cloud. When you're just starting the playbooks on your machine but they're executing stuff on `concent-builder`/`concent-deployment-server`, they're using the service accounts.
3. Remove service account keys from our secrets.
    - Also from secrets stored on any remote machines.
    - And ask other team members to update their secrets if they have them.
4. Update README.

bartoszbetka:
Resolves #372
It's depends on #378

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,30 +221,12 @@ cd concent-deployment/ethnode/
     --user      $user
 ```
 
-### Authorizing a user account on the build server to access the clusters
-This step must be performed separately for every user of the build server who needs to be able to access other parts of the project infrastructure on Google Cloud with `kubectl` or `gcloud`.
-It can be performed by user himself or an admin who can impersonate him with `sudo`.
+All the instructions below assume that you're using local playbooks to run build and deployment commands on a new server.
 
 The `$cluster` variable determines which server the playbook will be executed on.
 For `concent-dev` it connects to `concent-builder`.
 For other values (`concent-staging`, `concent-testnet` or `concent-mainnet`) - to `concent-deployment-server`.
 This behavior applies to all playbooks, except for `build-test-and-push-containers.yml`.
-
-The `$user_name` variable below indicates the user account to be authorized.
-To perform this step you need to have the .vault files with encrypted secrets in your local `concent-secrets/` directory.
-Only cloud secrets are required in this case.
-Ansible will prompt you for password required to decrypt them.
-
-```bash
-cd concent-deployment/cloud/
-ansible-playbook configure-user-authentication-for-clusters.yml    \
-    --extra-vars "cluster=$cluster user_name=$user_name"           \
-    --ask-vault-pass                                               \
-    --inventory  ../../concent-deployment-values/ansible_inventory \
-    --user       $user
-```
-
-All the instructions below assume that you're using local playbooks to run build and deployment commands on a new server.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -209,30 +209,12 @@ ansible-playbook create-vm-instances-for-geth.yml                   \
...
 ```
 
-### Authorizing a user account on the build server to access the clusters
-This step must be performed separately for every user of the build server who needs to be able to access other parts of the project infrastructure on Google Cloud with `kubectl` or `gcloud`.
-It can be performed by user himself or an admin who can impersonate him with `sudo`.
+All the instructions below assume that you're using local playbooks to run build and deployment commands on a new server.
 
 The `$cluster` variable decides about deployment to specific environment but also which server will be use for deployment process.
 If you want it for `mainnet`, `testnet` or `staging` environments you must create and configure the `concent-deployment` server.
 This behavior applies to all playbooks, except building container process (`build-test-and-push-containers.yml`).
```
cameel: Why is this now in the section about deploying the geth nodes?

I think your intention here was to remove this section completely but after a rebase some new lines jumped in. Please fix it.

bartoszbetka: I think that this part should stay. It informs about the assumptions of the instructions.

cameel: But it's in the wrong place. It's meant for someone interested in the sections below but you put it in a section about the geth client.

I would suggest this:
- Rename the `Deployment` section to `Initializing the deployment environments`
- I would start a new `Deployment` section just above `Building containers and cluster configuration`.
- Put the text that applies to all the subsections (starting from `All the instructions below assume ...`) at the beginning of that new section.

BTW, this text implies that there's a separate deployment server for each environment. Is this how it works now? I hope not.

bartoszbetka: It is an outdated change.
> BTW, this text implies that there's a separate deployment server for each environment. Is this how it works now? I hope not.

All suggestions are added in the properly PR: #382 . Please check it.

PR #382, opened 2019-05-10
Add information about privileges for service accounts

### LINKED ISSUE #368: Determine the minimal set of roles/privileges needed for the service accounts
Currently service accounts in our Google Cloud project have broad privileges. This would allow someone who managed to break into our build or deployment server to take control not only over that server but also over many aspects of the whole Google Cloud project. To limit these privileges we first need to find out which ones we actually need.

On the provisioned machines we'll need two levels of access to the project: _deployer access_ (can deploy and update a Concent cluster) and _cloud admin access_ (can add/delete parts of the infrastructure and also do everything a deployer can). Both of these should be able to do less than the built-in `Editor` role.

Basically, the deployer level should be enough for the build server (including Jenkins, implemented in #297). Cloud admin level is meant for the deployment server. These limitations are meant for actions that are automated via playbooks or shell scripts. Stuff that's not automated or needs to be performed from outside of Google Cloud infrastructure should rely on user's personal account.

Here are the specific actions we want to be able to perform at each of those levels:
- _Concent Deployer_:
    - Creating and deleting resources within a cluster (pods, services, config maps, etc.).
    - Creating and deleting schemas, tables and other items inside a PostgreSQL database on a Cloud SQL instance (but not the database or the instance itself).
    - Attaching storage to cluster pods.
    - Configuring a load balancer inside the cluster and attaching a static IP to it.
    - Pushing/pulling Docker images to/from Docker registry.
- _Concent Cloud Admin_:
    - Creating and deleting Google Compute Engine instances.
    - Creating and deleting Kubernetes clusters.
    - Creating and deleting Cloud SQL instances.
    - Creating and deleting PostgreSQL databases and roles in Cloud SQL.
    - Making Cloud SQL backups.
    - Creating and deleting disks.
    - Creating and deleting IPs.
    - Deleting images from Docker registry.

Among the things the cloud admin should **not** be allowed to do are:
- Changing roles/privileges.
    - But should be able to give service accounts access to specific clusters.
- Giving itself access to other Google Compute Engine instances.
- Using other service accounts without a key (note that the built-in `Editor` role can do that which is why we don't want the admin).

The goal of this task is to:
1. Determine which permissions or roles we need to implement these access levels. You can see most roles in [Understanding Roles](https://cloud.google.com/iam/docs/understanding-roles) but you may also need to take look at docs of specific Google Cloud features/components.
    - Predefined roles are fine even if they give a bit more privileges than we need as long as it's not unreasonable. If there's no predefined role that fits our needs, create a custom role (see [Creating and managing custom roles](https://cloud.google.com/iam/docs/creating-custom-roles)).
2. Document roles needed for each server in the README (or in a script/configuration file placed under `cloud/` in the repository).

bartoszbetka:
Resolves #368


README.md
```diff
@@ -2,6 +2,158 @@
...
+| `Cloud SQL Admin`          | Project                                       |
+| `Compute Firewall Admin`   | Project                                       |
+| `Service Account User`     | Default compute instance service account      |
+| `Service Account User`     | Concent Deployer's service account            |
+
+Commands below show how to create the account and assign the roles using command-line tools:
+
+``` bash
+gcloud beta iam service-accounts create \
+    "$deployer_service_account"         \
```
bartoszbetka: `$deployer_service_account` -> `$cloud_admin_service_account`

